PR #18, opened 2016-04-04
Added compatibility down to API 8

EmmanuelMess:


vanniktech:
Hmm I'm sorry but I'm quite hesitant here. I don't see any reason for supporting anything < API 15. I know it's not a big deal here on the library side but you should not really support all the way down to API 8.

EmmanuelMess:
@vanniktech Why is that? I'm creating an app that has compatibility with API 8 and I would like to add this library, and I'm sure backwards compatibility won't affect the library.

vanniktech:
Well we do need those other calls on lower apis so it does affect the library. Why do you need to support all the way down to 8? 15 minsdk is pretty common and new projects even use 17/18

EmmanuelMess:
@vanniktech "those other calls on lower apis" are ~14 lines and older phones are still in use and there should be compatibility with those too.

EmmanuelMess:
@vanniktech  "15 minsdk is pretty common and new projects even use 17/18", I didn't see any projects that are using higher APIs if they can use lower with code already in Android (no 3rd party compat lib).

vanniktech:
If you look at the [chart here](http://developer.android.com/about/dashboards/index.html) you can see that you really do not need minSdk 8. Also all of the other libraries out there have minsdk 15

EmmanuelMess:
@vanniktech I think 14 lines of code for 100% of the Play Store availability (instead of 95%) is worth it, if you don't, I have no more arguments to make, you can close the pull request.

EmmanuelMess:
@vanniktech I must say that if you don't pull this commit, I will have to think of some other way to add the functionality with compatibility down to API 8 into my app (you'll be credited accordingly and I wont be violating the license).

vanniktech:
Okay cool feel free to make a fork and publish the library with API 8.

Issue #75, opened 2017-01-06
Proposal: Switch to EmojiOne


rubengees:
This is a proposal for a large change in the library.

I think about switching to the emoji of [EmojiOne](http://emojione.com/). It is regularly updated, contains all of the latest emoji, has a friendly licence (creative commons) and provides mapping files for somewhat easy generation of mapping classes. 

### Benefits
- All of the missing emoji would be added #10, #11, #12.
- Integration of skin tones #34 (Not as a popup menu, but still better than not having them altogether).
- Looks also good on large displays 
- Nearly automatic updates possible, see down below.

### Caveats
- The library might grow in size, I'm not far enough to say that certainly.

### How I would approach this
- Write a small script for cloning the EmojiOne repo -> Copying the relevant resources -> Generate `.java` files which map to correct values. This has the advantage that we could just rerun the script when files of EmojiOne have changed and simply copy the changes to the library.
- Rewrite/Simplify the library to be compatible with the changes.

### Preview (Work in progress)
![device-2017-01-06-181100](https://cloud.githubusercontent.com/assets/8021265/21726470/1ce400b4-d43d-11e6-8819-4131f516a669.png)

What do you think about this? I would happily submit a PR for that (in fact I'm already working on it, the script is nearly ready).

vanniktech:
What do you think of providing support for both (maybe as a separate artifact)?

rubengees:
Hmm, I don't know if that would be (easily) possible as the current approach and the new one would be somewhat different and hard to to abstract into one solution. But I will look into it.

rubengees:
On second thought: Why would you like to keep the old ones? For backward support? As I noticed nobody is using the old iOS ones anymore (WhatsApp, Telegram, ...).

vanniktech:
If so they'd need to be different artifacts hence also different library modules. 

WhatsApp is now on Android using the newer iOS emojis which are also the ones this library should provide.

rubengees:
I thought about something like a commons module with contains all the ui and managing part. Then there would be different modules, which the user could choose from, e.g. old-ios, emojione, android, whatever. But as I said I think that might be quite hard to get that working. Especially since the current emojis are mapped by hand. 

Ah, those are new iOS ones? Haven't used iOS for ages...
I looked for them, but couldn't find a repo or something parsable containing them.

vanniktech:
Yeah I also didn't find them. Although I didn't look for them either. Common module might make sense although there's not that much common code.

[...]

PR #77, opened 2017-01-10
Rewrite for more Emojis, modularity and performance

rubengees:
As discussed in #75.

#### Generator

The folder `generator` contains a NodeJS script for updating the images and generating relevant java classes.
This is done like follows: Download the file containing the emoji and the mapping file -> Index all the emojis and categories -> Copy the images to the `drawable` folders -> Generate the java files and copy them directly into the respective folders.

Emojis are downloaded from [here](http://unicode.org/emoji/charts/full-emoji-list.html) and the mapping file is from [here](https://github.com/github/gemoji).

I may add a README with instructions on how to run the script later.

#### Emojis

This adds the latest iOS and EmojiOne Emojis. There are different modules for different variants.

The benefits are:
- Complete collection of all emojis (even with skin tones and even more than WhatsApp Beta currently has!)
- Better image quality
- Choice of Emoji collection

#### Modularization

There is now a core module named "emoji". For each emoji variant there is a "emoji-\<variant\>" module.
As of this PR this are "emoji-ios" and "emoji-one".
These modules have a dependency on "emoji", so the user can just write `compile com.vanniktech.emoji-ios:version` for instance.
Furthermore, the user can pass an own instance of the new interface `EmojiProvider`. It is very easy to implement (just one method).

The `EmojiProvider` is then (and has to be) registrered like this (In an `Application` subclass for example):
```java
EmojiManager.install(new IOSEmojiProvider()); // Or EmojiOneProvider or 
                                              // custom implementation
```

#### Categories

Categories are now also generated and thus need to be more abstract. For this purpose, a new interface `EmojiCategory` which holds the icon for the category and all the emojis in the category. The `EmojiView` was changed to work with this new approach.
The icons of the categories are from EmojiOne (including the `recent` icon), except the backspace icon, which I generated with the Android Studio Asset Studio. All of these icons are now vectors.

##### Parsing

- I rewrote large parts of the emoji parsing to be much more flexible. I added a new data structure `EmojiTree` which holds paths to each emoji (each child is one character, if the child is the `end` of an emoji, it holds the respective resource). When parsing, the tree is traversed and the longest fitting path is used. This idea is from [emoji-java](https://github.com/vdurmont/emoji-java), which does it quite similar.

- A new singleton class `EmojiManager` holds the `EmojiTree` (and the categories), which is filled upon construction (Like the `EmojiHandler` before).
- This also greatly simplifies the `EmojiHandler`, which now only iterates the String and looks up the emojis.

##### Category adapter

I noticed that the emoji picker is a bit laggy. To help with that, the `EmojiArrayAdapter` was changed to asynchronously load the images. This is done with a plain `AsyncTask`. Moreover the current approach with TextViews were replaced by ImageViews as this further improves performance.
Another problem was, that the `GridViews` were hold in memory even when the `PagerAdapter` called the `destroy`-method. This could lead to OOMs, especially since the images know have better quality and thus are larger. This is know changed so that the `GridView` is inflated when requested and also properly destroyed.

##### Emoji class

I changed the Emoji class a bit to better work with the new architecture. It know holds the associated drawable resource and has new constructors. This required the `RecentEmojiManager` implementation to change a bit. Upon loading from disk, it searches the `EmojiManager` for the existing unicode. If not found, the emoji is discarded (This also improves compability between modules. If the user changes for example from ios to emojione and the latter does not contain an Emoji, the App will not crash).
Furthermore it is also used in the new `EmojiTree`, reusing existing instances and thus minimizing unnecessary object allocation like before (Unicode Strings were allocated both for categories and the `EmojiHandler`).

##### New Views

I added a new View: SquareImageView`. As the name suggests, it draws the image in a square, with the width set for the height. This is used in the emoji picker.

##### Minor other changes

- I altered the value for the `EmojiGridView` item width a bit to make it look good with the new images.

**Just to clarify again: These changes are not directly backwards compatible**

### This is what it looks like 

iOS | EmojiOne
-- | --
device-2017-01-29-165256 | device-2017-01-29-165515

### Problems I noticed

This is not directly related to this PR, but I noticed it when I worked on it:  
When adding many emojis to the `EmojiEditText` it becomes quite slow (I measured once and managed to get a freeze of 500ms). This is due to the `EmojiHandler` adding a span for each emoji, which is very unperformant. I looked into that, but didn't found a better solution... Maybe you now of a different way to approach this?

I am almost certain that I forgot something so please have a look yourself (Especially at the `Gradle` config, I'm not sure if it is right and have no way of testing)!

Moreover I do apologize for the awfully large PR, which makes GitHub freeze :grin:.


vanniktech:
This will replace the current emojis right? If so that's not really doable / acceptable since a lot of the library consumers rely on that particular style.

rubengees:
Yes, this would replace the current ones. I wanted to modularize it in another commit, as this one is already very large, but I can work on it in this one if you prefer.

vanniktech:
I've been mulling over this and I'd accept it if it's in a separate module and the current version isn't touched. So that you can either use:

`compile 'com.vanniktech:emoji:0.3.0'`

or 

`compile 'com.vanniktech:emoji-one:0.3.0'`

and they have similiar APIs / usage.

Of course I can also understand it if you want to fork this library and maintain your version for emojis with EmojiOne.

rubengees:
I thought about it and would do it like this:

One core module, which has the `EmojiTextView`, `EmojiEditText` and managing functions. This would then be `compile 'com.vanniktech:emoji:0.5.0' ` (Or whatever version it is then).
The core module would need some entry point for installing the chosen emojis, maybe in the `Application` or for each `EmojiTextView` and `EmojiEditText`.
I could imagine the first looking like this:

```java
Emoji.install(new EmojiOneProvider()); // Or new IOsProvider()
```

The second would look like this:

```
emojiEditText.setEmojiProvider(new EmojiOneProvider());
```

Personally I like the first one better, as we would have a singleton then, so we don't hold duplicate data in the memory.  
What do you think?

We would then have modules for each variant.
This would be `compile 'com.vanniktech:emoji-one:0.3.0'` or `compile 'com.vanniktech:emoji-ios:0.3.0'` for example. Each of these modules would have a transitive dependency on the core module so you can just write what variant you would like to have.  
Yes, this requires the user to change it's dependency once and adjust usage a bit, but I think that is better for modularity and clearnes in the long run. Moreover it has the benefit, that the user **could** provide its own `EmojiProvider` or even something entirely different (The new architecture allows to replace any text by images).

vanniktech:
Doing something via a provider sounds good to me. Also I don't care that much about backwards compatibility at this point since we're still not v1. Plus if the package names are changed, just upgrading won't work since the current package won't be available with then next version.

rubengees:
I have added modularization, finally found a source for all emojis and a view other things. Please have a look at the updated description!

I have no idea why this one test fails, apart from this everything (checkstyle, PMD, findbugs, lint) works. Do you know what is causing this?

Also especially have a look on the `Gradle` config and how the project is structured, as it don't know if this is correct and I have now way to test it.

.travis.yml
```diff
@@ -34,7 +34,7 @@ branches:
 notifications:
     email: false
 
-sudo: false
```
vanniktech: how come this needs to be changed?

app/build.gradle
```diff
@@ -52,7 +54,7 @@ android {
 }
 
 dependencies {
-    compile project(':library')
+    compile project(':emoji-ios')
```
vanniktech: either we need a 2nd sample or include both into one sample

emoji-ios/src/main/java/com/vanniktech/emoji/ios/IOSEmojiProvider.java
```diff
@@ -0,0 +1,36 @@
...
+import com.vanniktech.emoji.ios.category.NatureCategory;
+import com.vanniktech.emoji.ios.category.ObjectsCategory;
+import com.vanniktech.emoji.ios.category.PeopleCategory;
+import com.vanniktech.emoji.ios.category.PlacesCategory;
+import com.vanniktech.emoji.ios.category.SymbolsCategory;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+public class IOSEmojiProvider implements EmojiProvider {
```
vanniktech: final class

emoji-ios/src/main/java/com/vanniktech/emoji/ios/IOSEmojiProvider.java
```diff
@@ -0,0 +1,36 @@
...
+    @Override
+    @NonNull
+    public Map<String, EmojiCategory> getCategories() {
+        final LinkedHashMap<String, EmojiCategory> result = new LinkedHashMap<>();
```
vanniktech: type declaration should be Map

emoji-ios/src/main/java/com/vanniktech/emoji/ios/IOSEmojiProvider.java
```diff
@@ -0,0 +1,36 @@
...
+
+        result.put("People", new PeopleCategory());
```
vanniktech: Should not those values be defined somewhere? (also are they used for translations?)

.travis.yml
vanniktech: Okay I just self answered it when I read your messages again :D

vanniktech:
> This is not directly related to this PR, but I noticed it when I worked on it:
When adding many emojis to the EmojiEditText it becomes quite slow (I measured once and managed to get a freeze of 500ms). This is due to the EmojiHandler adding a span for each emoji, which is very unperformant. I looked into that, but didn't found a better solution... Maybe you now of a different way to approach this?

I don't. At least not right now.

Is it possible to get the old categories icons back for the emjoji-ios?

I'll also go over this one here and might address my issues that I pointed out.

emoji-ios/src/main/java/com/vanniktech/emoji/ios/IOSEmojiProvider.java
rubengees: Hmm, they are not really used anywhere, I think it's best to make `getCategories` just return a `List`.

app/build.gradle
rubengees: How about an entry in the toolbar for switching the available emoji variants?

rubengees:
> Is it possible to get the old categories icons back for the emjoji-ios?

Yes, but they have bad quality so I dropped them in favour of the ones from EmojiOne. I think the category icons can be used for all variants as they do not have a very specific style.

vanniktech:
ops @rubengees I just created #79 so I can add my changes too can you try to get your latest commit `9813f0b` in there?

vanniktech:
Oh cool you put my changes back in. I'll close the other one then.

rubengees:
I'm sorry, I'm not really good with git...

vanniktech:
Don't mind, we'll somehow make it work :)

vanniktech:
Getting:

```
/Users/nik/tmp/Emoji/emoji-ios/src/main/java/com/vanniktech/emoji/ios/category/ActivityCategory.java:112: error: cannot find symbol
            new Emoji(new int[]{0x1f6b5, 0x200d, 0x2640, 0xfe0f}, R.drawable.emoji_ios_1f6b5_200d_2640_fe0f),
```

rubengees:
Ah! Sorry, I forgot to actually add the images :sweat_smile:

vanniktech:
Nice looking better and better, I also updated the issue

rubengees:
Just pushed the change for the first point.
The second point make some time as I want to search for better quality ones (Hopefully even vectors). If I do not find some, I'll include the old ones.

vanniktech:
@rubengees why not simply use the old icons for now? There's no harm in that we can change them later if needed.

rubengees:
I have new ones. They are PNGs and don't have the best quality, but better than before.

vanniktech:
Awesome. Thank you so much for all of your work. Tomorrow I'm on a train and I'll do some more adjustments locally to the code and prepare some more things. It really looks fabulous. Either I'll release 0.4.0 tomorrow or once I'm back from vacation (~8 days). SNAPSHOT version will be built automatically and should be available soon after the merge.

rubengees:
Great, we did it :clap:
Looking forward to 0.4!

vanniktech:
All of the new ios emojis are right now adding an extra of 3MB :/ on top of the 2 or 3 MB it was before. 

I also tried compressing them further but had no luck. I guess that's just the drawback of supporting more emojis. Another option might be vector drawables although I'm not sure you can get emojis in that format. Do you know of anything @rubengees ?

rubengees:
@vanniktech Yes, thats the price for having them all...

Only EmojiOne has vectors as far as I know. I haven't found them for other variants at least.

vanniktech:
Tracking in #88

PR #93, opened 2017-02-21
Make the EmojiEditText more performant

rubengees:
The problem was the recalculation of all spans when changing the text.
With this change, the existing spans are reused and only new text is taken into account.
Furthermore the flag `SPAN_EXCLUSIVE_EXCLUSIVE` is now in use, which causes spans with the length 0 (When an emoji has been deleted) to be automatically removed by the platform.
This also reduces the calls to `findEmoji`, which can add up to bad performance.

The problem still persists when text is inserted at the start of the `Spannable` though, as the spans get recalculated by the platform then, but there is sadly nothing we can do about that.

### DIFF
--- a/emoji/src/main/java/com/vanniktech/emoji/EmojiHandler.java
+++ b/emoji/src/main/java/com/vanniktech/emoji/EmojiHandler.java
@@ -3,9 +3,7 @@ package com.vanniktech.emoji;
 import android.content.Context;
 import android.support.v4.util.Pair;
 import android.text.Spannable;
-
 import com.vanniktech.emoji.emoji.Emoji;
-
 import java.util.ArrayList;
 import java.util.List;
 
--- a/emoji/src/main/java/com/vanniktech/emoji/EmojiHandler.java
+++ b/emoji/src/main/java/com/vanniktech/emoji/EmojiHandler.java
@@ -80,13 +80,5 @@ final class EmojiHandler {
       this.start = start;
       this.end = end;
     }
-
-    int getStart() {
-      return start;
-    }
-
-    int getEnd() {
-      return end;
-    }
   }
 }
--- a/emoji/src/main/java/com/vanniktech/emoji/EmojiHandler.java
+++ b/emoji/src/main/java/com/vanniktech/emoji/EmojiHandler.java
@@ -1,7 +1,6 @@
 package com.vanniktech.emoji;
 
 import android.content.Context;
-import android.support.v4.util.Pair;
 import android.text.Spannable;
 import com.vanniktech.emoji.emoji.Emoji;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


emoji/src/main/java/com/vanniktech/emoji/EmojiHandler.java
```diff
@@ -1,34 +1,75 @@
...
       }
     }
   }
 
-  private EmojiHandler() {
-    throw new AssertionError("No instances.");
+  static final class SpanRangeList {
+    static final int SPAN_NOT_FOUND = -1;
+
+    private final List<Pair<Integer, Integer>> spanRanges = new ArrayList<>();
```
vanniktech: Might be better to write our own Pair class that uses primitive `int` instead of `Integer` otherwise you'll have plenty of boxing and unboxing which can also cause bad performance

emoji/src/main/java/com/vanniktech/emoji/EmojiHandler.java
```diff
@@ -1,34 +1,75 @@
...
 import com.vanniktech.emoji.emoji.Emoji;
+import java.util.ArrayList;
+import java.util.List;
+
+import static com.vanniktech.emoji.EmojiHandler.SpanRangeList.SPAN_NOT_FOUND;
 
 final class EmojiHandler {
-  static void addEmojis(final Context context, final Spannable text, final int emojiSize) {
-    final EmojiSpan[] spans = text.getSpans(0, text.length(), EmojiSpan.class);
+  private EmojiHandler() {
```
vanniktech: nit: put it between static method addEmojis and the class

emoji/src/main/java/com/vanniktech/emoji/EmojiHandler.java
rubengees: Good idea! Done.

rubengees:
Everything done.

emoji/src/main/java/com/vanniktech/emoji/EmojiHandler.java
```diff
@@ -41,35 +36,57 @@ static void addEmojis(final Context context, final Spannable text, final int emo
...
+  static final class Range {
+    final int start;
+    final int end;
+
+    Range(final int start, final int end) {
+      this.start = start;
+      this.end = end;
+    }
+
+    int getStart() {
```
vanniktech: you don't need those just access `start` directly

emoji/src/main/java/com/vanniktech/emoji/EmojiHandler.java
```diff
@@ -41,35 +36,57 @@ static void addEmojis(final Context context, final Spannable text, final int emo
...
+      return start;
+    }
+
+    int getEnd() {
```
vanniktech: you don't need those just access `end` directly

PR #94, opened 2017-02-21
Add an editorconfig file

rubengees:
This adds an [editorconfig file](http://editorconfig.org/) to be used by Android Studio and others.
It includes the indentation of two spaces and some general adjustments like setting the charset to `utf-8`.


.editorconfig
```diff
@@ -0,0 +1,15 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.md]
+trim_trailing_whitespace = false
```
vanniktech: true

rubengees:
I thought it would be a good way for contributors to conform to your code style without changing the Android Studio settings manually (I had some problems getting Android Studio to not auto-format the code), but it's okay if you don't want it in.

vanniktech:
Decided against it.

PR #95, opened 2017-02-22
Add support for Google emojis

rubengees:
This adds support for google emojis by introducting the new module `emoji-google`.
I added the target to the `generator` and also updated the sample app for it.

One noteworthy change is the possibility to ignore individual emojis when generating. This was needed, because Google had some weird question mark flags, which I added to the new ignore list by hand.


rubengees:
We could also have `Twitter`, `Facebook`, `Facebook Messenger` (Those are different), `Samsung` and `Windows`.
I don't know if all of those make sense though, since they are not that widely used. What do you think we should also include?

rubengees:
You can find a list of the different variants [here](http://unicode.org/emoji/charts/full-emoji-list.html). This is the file we use for parsing, so exactly those images would be in the lib.

vanniktech:
Hmm the only thing that might make sense to me personally is the facebook messer emojis. But then again if it's for free, why not provide all of them?

rubengees:
Yes, thats true. I will work on another PR.

Issue #96, opened 2017-02-24
Investigate better Popup logic


rubengees:
- Version of the library: 0.5
- Affected devices: All
- Affected versions: All

The current logic for showing the picker itself and the skin tone popup have some problems.
This includes:
- Does not work with physical keyboards, because the `OnGlobalLayoutListener` does not get called (#31).
- Problems with dialogs and other contexts, apart from a normal full screen layout (#26).
- Skin tone popup does not behave exactly like it should be. If you do a long click, but dont lift your finger, you are still able to move the list. Moreover does an outside click counts as a real click. If you click outside the popup on an emoji, the popup is closed **and** the emoji is selected, while it should only close the popup.
- The picker does also not behave exactly like it should be. Other pickers like this close when clicked outside, but it is arguable if that is what we want.
- The popup closes when the `Activity` is paused, or orientation is changed. Other pickers like this do remain open
- Risky code
  - The `MIN_KEYBOARD_HEIGHT` is hardcoded to 100 pixels. I doubt that there are keyboards smaller than that, but it is still risky.
  - Usage of the Android internal field `status_bar_height`, which could change or disappear in future versions.

I tested around a bit, but couldn't really get to anything. One notable thing though is the following:
To make a popup window behave properly (Like I described above for the skin tone popup), one can set:

```java
popupWindow.setFocusable(true);
popupWindow.setOutsideTouchable(true);
```

This resulted in this weird behaviour though:

![unbenannt](https://cloud.githubusercontent.com/assets/8021265/23287097/987bba7c-fa3b-11e6-9444-a87474290008.png)

I don't really know how to approach this. Do you know something @vanniktech? I will further investigate this though.

vanniktech:
> The picker does also not behave exactly like it should be. Other pickers like this close when clicked outside, but it is arguable if that is what we want.

I think the current behavior is what we want. 

> The popup closes when the Activity is paused, or orientation is changed. Other pickers like this do remain open

That might be the case. In 99% of my apps I only support portrait mode and do not really care about this behavior that much.

> The MIN_KEYBOARD_HEIGHT is hardcoded to 100 pixels. I doubt that there are keyboards smaller than that, but it is still risky.

Yup but that has always been the workaround for detecting keyboards and was The Solution™

> Usage of the Android internal field status_bar_height, which could change or disappear in future versions.

Also for that I'm not aware of another solution.

```java
popupWindow.setFocusable(true);
popupWindow.setOutsideTouchable(true);
```

[...]

PR #97, opened 2017-02-25
Improve Popup logic

### LINKED ISSUE #26: shows above keyboard in dialog box
I have a dialog box in which user can enter text and emoji.
But when i open emoji panel it shows above the keyboard instead of top of the keyboard.
I have given root view of the dialog box custom xml root view

rubengees:
This would address some of the points in #96 and fix #26.

### Contents of this Pull Request

The internal popup logic has been improved in several ways:

- Support for dialogs and other places apart from a normal `Activity`.
- Better behaviour of variant popups. If the user opens one and clicks outside of it, the event is not delegated, but only the popup is closed.
- Better keyboard height detection
- Fixes for identification if the keyboard is open or not. Previously, when one opened the picker, clicked the toggle item and closed the keyboard by pressing the back button, two clicks were required to open the popup again, as the `EmojiPopup` had a wrong state concerning `isKeyboardOpen`. 
- The `onEmojiPopupShownListener` is only shown, after the popup is really shown. Previously it was called, when we started opening the keyboard, which could lead to a wrong state if it was not shown (If a hardware keyboard is present for example).

#### Code changes to achive this

- The height of the current `rootView` is calculated from the `DecorView`.
- Better places for setting and removing the global layout listener were chosen.
- The popup is always shown at the bottom of the screen, following the advice in #26.
- `popupWindow.setFocusable(true)` is now called on the variant popup. `popupWindow.setInputMethodMode(PopupWindow.INPUT_METHOD_NOT_NEEDED)` was the key to make that work without weird side effects.
- Adjusted location calculation for the variant popup, to make it work in dialogs.

I have added a simple Dialog to the sample for showcasing and testing.

#### Things concerning backwards compability

- The popup now closes itself when the keyboard is closed. Thus, a call to `emojiPopup.dismiss();` in `onKeyboardClose` is not necessary anymore.
- We require the `Context` of the `rootView` to be an `Activity` now. We need that as we want to access the `Window` for better height calculation.

### DIFF
--- a/emoji/src/main/java/com/vanniktech/emoji/EmojiVariantPopup.java
+++ b/emoji/src/main/java/com/vanniktech/emoji/EmojiVariantPopup.java
@@ -1,7 +1,8 @@
 package com.vanniktech.emoji;
 
-import android.app.Activity;
+import android.content.Context;
 import android.graphics.Bitmap;
+import android.graphics.Point;
 import android.graphics.drawable.BitmapDrawable;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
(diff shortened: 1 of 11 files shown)


emoji/src/main/java/com/vanniktech/emoji/EmojiVariantPopup.java
```diff
@@ -21,38 +22,80 @@
...
+        // If the actual location the popup is shown at differs from the wanted (Can happen with
+        // dialogs), we have to fix it.
+        if (!(actualLocation.x == x && actualLocation.y == y)) {
+          final int differenceX = actualLocation.x - x;
+          final int differenceY = actualLocation.y - y;
+
+          final int fixedOffsetX = actualLocation.x > x ? x - differenceX : x + differenceX;
+          final int fixedOffsetY = actualLocation.y > y ? y - differenceY : y + differenceY;
+
+          popupWindow.update(fixedOffsetX, fixedOffsetY, DONT_UPDATE_FLAG, DONT_UPDATE_FLAG);
```
rubengees: This might look weird, but is needed for dialogs. I have not found a way to calculate the offset properly, so we `fix` it here if it differs from the wanted location.

rubengees:
I assume that happens because of the hardware keyboard, Genymotion uses.
Did it work before (It shouldn't...)?

vanniktech:
Yup before it worked perfectly. Also I have the soft keyboard activated and it shows up.

rubengees:
Okay, I'll investigate. On which API Level does it happen?

vanniktech:
Genymotion Google Nexus 5X 7.0.0 API 24

rubengees:
Fixed in bf2fc8b. This seemed to be a [bug](http://stackoverflow.com/questions/30827576/popupwindow-overlaps-soft-buttons-on-android-5-0/30868419) on APIs greater than 21. We now "fix" the location, exactly like the variant popup.

Furthermore I made the `Context` to `Activity` cast from the `rootView` more safe (The [support lib](http://grepcode.com/file/repository.grepcode.com/java/ext/com.google.android/android/4.3_r1/android/support/v7/app/MediaRouteButton.java#MediaRouteButton.getActivity%28%29) does it in the same way) .

emoji/src/main/java/com/vanniktech/emoji/Utils.java
```diff
@@ -62,6 +63,20 @@ static int screenHeight(@NonNull final Activity context) {
     return result;
   }
 
+  static Activity contextToActivity(@NonNull final Context context) {
```
vanniktech: method name: asActivity

PR #108, opened 2017-04-04
Remove right Scrollbar in Categories

RicoChr:
I succesfully removed the scrollbars programmatically


RicoChr:
Didn't quite understand the concept of codecov, so if I need to correct something, let me know.

vanniktech:
Thanks !

PR #118, opened 2017-05-08
Adjust generator for emoji 5.0

rubengees:
This PR makes the `generator` script ready for the updated emoji 5.0 files.
This was necessary, as the format of the webpage we are parsing changed.

Another change is, that we now use the `emoji.json` from EmojiOne for finding meta info.
Moreover I added some more codes to ignore for Google, since they added more "?"-Flags.

I will create another PR for the new images to keep this one simple.


vanniktech:
Thanks !

PR #119, opened 2017-05-08
Update all emojis to emoji 5.0

rubengees:
This adds some flags and various symbols to `Ios` and `Google`. `EmojiOne` gets a totally new set, as they recently released [3.0](https://www.emojione.com/blog/introducing-emojione-v30-at).

#### Problems

- This PR is way to large 😁
- There are duplicate flags now. This happens for example for the United States flag and the U.S. Outlying Islands flag, which share the same image. There are other examples as well. Do we care? I couldn't think of a way of fixing this, without excluding them manually.


vanniktech:
@rubengees can you update your branch and make sure that it's all set up?

vanniktech:
One last thing, can you in the generator where the `ignore` constant is used specify what kind of emojis those are (basically that those are duplicate flags).

rubengees:
Done! I noticed that `lint` complains about duplicate icons which made it easy to ignore them.

vanniktech:
Yup, I also saw it. Thanks again for this and I'll merge it as soon as Travis is done building this PR.

Issue #128, opened 2017-05-31
API to query Emoji information about a String


vanniktech:
It'd be cool if you could query information about a String, such as whether it consists of only one Emoji or whether the entire String consists of multiple Emojis. Another one would be if the String consists of a specific Emoji (such as the heart ❤️ ), so it could be animated.

One use case would be to display the Emoji in a bigger size if it's only one. And for the multiple ones they might get a bit bigger too.

rubengees:
This will be possible to some intent with the change i am working on.

vanniktech:
Yup. I think the Tree will need to be reversed. That way also custom Emojis / provided Emojis would work.

vanniktech:
Closed by #145

PR #129, opened 2017-05-31
Performance optimization

### LINKED ISSUE #127: EmojiHandler.addEmojis is horribly slow
Hi! First of all, thanks for this library! I am currently using version 0.5.0-SNAPSHOT (due to issue #103), I don't know if this only happens with this version.

I have seen that `EmojiTextView` and similar are slow when the text is several lines long (I am using a text of about 10000 characters as a test). This blocks the UI thread and is specially noticeable when scrolling a `RecyclerView`.

When running on an emulator (with API 23 to be precise), executing `setText` on an `EmojiTextView` takes about half a second. When doing that in a normal `TextView` the result is immediate (less than 0.001 seconds). I have tried to analyze the behavior of the library and the problematic method is `EmojiHandler.addEmojis`, that seems to parse the full string in a while loop.

I have changed the code in my case to use a regular expression, and the result is amazing: it now takes around 0.002 seconds to execute setText, and the behavior is the same.

Even if you don't take the regexp approach, that method should be checked for optimization.

I attach a Gist of my changed class: https://gist.github.com/Ereza/08282caa4fa444bcbe41aedcc1c820b2

rubengees:
This implements and fixes #127.

### Contents of this Pull Request

Looking up emojis in the `EmojiTextView` and `EmojiEditText` should now be more performant.
I switched to the regex-based apporach, proposed by @Ereza. I noticed bad performance in the `EmojiEditText` when there are many emojis, so I mixed the best of both approaches.
There should be no difference in behaviour.

### Code changes to achieve this

- We create a regex `Pattern` now during the `EmojiManager#install` method.
- I nuked the entire `EmojiTree` and replaced it with a simple `Map`. This is possible, as the regex now acts as our `EmojiTree`. This should be faster, since we now do not need to traverse multiple layers of data structures. It should also be more memory efficient.
-  I moved all finding-logic to the `EmojiManager`. The benefits are:
    - It semantically fits better there imo.
    - We do not need to expose the `Pattern` or other internal logic.
    - We may make the api of the `EmojiManager` public for other use cases. It is able to find emojis in a `CharSequence` and returns the range they are located in. This would also implement most of #128.
- The `EmojiHandler` still has logic for reusing existing `Spans`. This is the difference to the Gist of @Ereza.
- The `EmojiManager#install` method now requires to install an `EmojiProvider` with at least one `EmojiCategory`, having at least one `Emoji`.

I also upgraded `Gradle`, the plugin and the `errorprone` plugin, to make it run in my `Android Studio` 3.0 installation. I hope that's fine. I also had to fix some lint issues which popped up.


emoji/src/main/java/com/vanniktech/emoji/EmojiHandler.java
```diff
@@ -2,83 +2,28 @@
...
+import com.vanniktech.emoji.EmojiManager.EmojiRange;
 import java.util.ArrayList;
 import java.util.List;
 
-import static com.vanniktech.emoji.EmojiHandler.SpanRangeList.SPAN_NOT_FOUND;
-
 final class EmojiHandler {
-  static void addEmojis(final Context context, final Spannable text, final int emojiSize) {
-    final SpanRangeList existingSpanRanges = new SpanRangeList(text);
+  static void replaceWithImages(final Context context, final Spannable text, final int emojiSize) {
```
rubengees: Sneaking this name change in, I hope that's fine.

build.gradle
```diff
@@ -20,9 +20,9 @@ buildscript {
         classpath 'com.vanniktech:gradle-code-quality-tools-plugin:0.4.0'
         classpath 'com.vanniktech:gradle-android-junit-jacoco-plugin:0.4.0'
 
-        classpath 'com.android.tools.build:gradle:2.2.2'
+        classpath 'com.android.tools.build:gradle:2.3.2'
         classpath 'com.github.ben-manes:gradle-versions-plugin:0.14.0'
```
vanniktech: Then also feel free to bump this one to `0.15.0`

build.gradle
```diff
@@ -20,9 +20,9 @@ buildscript {
         classpath 'com.vanniktech:gradle-code-quality-tools-plugin:0.4.0'
         classpath 'com.vanniktech:gradle-android-junit-jacoco-plugin:0.4.0'
```
vanniktech: `0.6.0`

emoji/src/main/java/com/vanniktech/emoji/EmojiHandler.java
```diff
@@ -2,83 +2,28 @@
...
-    int index = 0;
-
-    while (index < text.length()) {
-      final int existingSpanEnd = existingSpanRanges.spanEnd(index);
-
-      if (existingSpanEnd == SPAN_NOT_FOUND) {
-        final int nextSpanStart = existingSpanRanges.nextSpanStart(index);
-        final int searchRange = nextSpanStart == SPAN_NOT_FOUND ? text.length() : nextSpanStart;
-        final Emoji found = emojiManager.findEmoji(text.subSequence(index, searchRange));
+    final List<Integer> existingSpans = new ArrayList<>();
```
vanniktech: extract `text.getSpans(0, text.length(), EmojiSpan.class)` and then use the size to initialize the List so that it won't increase it's internal limit while adding items to reduce allocations

emoji/src/main/java/com/vanniktech/emoji/EmojiHandler.java
```diff
@@ -2,83 +2,28 @@
...
-          index += found.getLength();
-        } else {
-          index++;
-        }
-      } else {
-        index += existingSpanEnd - index;
+    for (final EmojiRange location : emojiManager.findAllEmojis(text)) {
+      if (!existingSpans.contains(location.getStart())) {
+        text.setSpan(new EmojiSpan(context, location.getEmoji().getResource(), emojiSize),
+                location.getStart(), location.getEnd(), Spannable.SPAN_EXCLUSIVE_EXCLUSIVE);
```
vanniktech: you can move all into one line

emoji/src/main/java/com/vanniktech/emoji/EmojiHandler.java
```diff
@@ -2,83 +2,28 @@
...
+    }
 
-          index += found.getLength();
-        } else {
-          index++;
-        }
-      } else {
-        index += existingSpanEnd - index;
+    for (final EmojiRange location : emojiManager.findAllEmojis(text)) {
+      if (!existingSpans.contains(location.getStart())) {
```
vanniktech: exract `location.getStart()` to save a call

emoji/src/main/java/com/vanniktech/emoji/EmojiManager.java
```diff
@@ -34,32 +50,98 @@ static EmojiManager getInstance() {
...
 
       //noinspection ForLoopReplaceableByForEach
       for (int j = 0; j < emojis.length; j++) {
-        INSTANCE.emojiTree.add(emojis[j]);
+        final Emoji currentEmoji = emojis[j];
+
+        INSTANCE.emojiMap.put(currentEmoji.getUnicode(), currentEmoji);
+        unicodesForPattern.add(currentEmoji.getUnicode());
+
+        for (final Emoji variant : currentEmoji.getVariants()) {
```
vanniktech: also fori to reduce iterator allocation?

emoji/src/main/java/com/vanniktech/emoji/EmojiManager.java
```diff
@@ -34,32 +50,98 @@ static EmojiManager getInstance() {
...
     //noinspection ForLoopReplaceableByForEach
     for (int i = 0; i < INSTANCE.categories.length; i++) {
       final Emoji[] emojis = checkNotNull(INSTANCE.categories[i].getEmojis(), "emojies == null");
 
       //noinspection ForLoopReplaceableByForEach
       for (int j = 0; j < emojis.length; j++) {
-        INSTANCE.emojiTree.add(emojis[j]);
+        final Emoji currentEmoji = emojis[j];
+
+        INSTANCE.emojiMap.put(currentEmoji.getUnicode(), currentEmoji);
```
vanniktech: extract `currentEmoji.getUnicode()`

emoji/src/main/java/com/vanniktech/emoji/EmojiManager.java
```diff
@@ -34,32 +50,98 @@ static EmojiManager getInstance() {
...
+        unicodesForPattern.add(currentEmoji.getUnicode());
+
+        for (final Emoji variant : currentEmoji.getVariants()) {
+          INSTANCE.emojiMap.put(variant.getUnicode(), variant);
```
vanniktech: extract `variant.getUnicode()`

emoji/src/main/java/com/vanniktech/emoji/EmojiManager.java
```diff
@@ -34,32 +50,98 @@ static EmojiManager getInstance() {
...
+          unicodesForPattern.add(variant.getUnicode());
+        }
       }
     }
+
+    if (unicodesForPattern.isEmpty()) {
+      throw new IllegalArgumentException("Your EmojiProvider must at least have one category with at least one emoji");
```
vanniktech: nit: end the sentence with a dot

emoji/src/main/java/com/vanniktech/emoji/EmojiManager.java
```diff
@@ -34,32 +50,98 @@ static EmojiManager getInstance() {
...
   }
 
   void verifyInstalled() {
     if (categories == null) {
       throw new IllegalStateException("Please install an EmojiProvider through the EmojiManager.install() method first.");
     }
   }
+
+  static class EmojiRange {
+    private final int start;
```
vanniktech: remove private modifier and the get methods

emoji/src/main/java/com/vanniktech/emoji/EmojiManager.java
```diff
@@ -34,32 +50,98 @@ static EmojiManager getInstance() {
...
+    private final int end;
+    private final Emoji emoji;
```
vanniktech: remove private modifier and the get methods

emoji/src/main/java/com/vanniktech/emoji/EmojiManager.java
```diff
@@ -34,32 +50,98 @@ static EmojiManager getInstance() {
...
 
   void verifyInstalled() {
     if (categories == null) {
       throw new IllegalStateException("Please install an EmojiProvider through the EmojiManager.install() method first.");
     }
   }
+
+  static class EmojiRange {
+    private final int start;
+    private final int end;
```
vanniktech: remove private modifier and the get methods

emoji/src/main/res/values/library_emoji_strings.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<resources>
```
rubengees: Lint complained about unused resources here, failing the build.

emoji/src/main/res/values/library_emoji_strings.xml
```diff
@@ -1,5 +1,5 @@
...
+<resources xmlns:tools="http://schemas.android.com/tools" tools:ignore="UnusedResources">
```
vanniktech: Hmm this should be already configured via a lint.xml file

emoji/src/main/java/com/vanniktech/emoji/RecentEmojiManager.java
```diff
@@ -133,7 +133,7 @@ int size() {
       return emojis.size();
     }
 
-    @Override public Iterator<Data> iterator() {
```
rubengees: Also, lint was complaining about this one.

emoji/src/test/java/com/vanniktech/emoji/EmojiManagerTest.java
```diff
@@ -49,10 +47,26 @@
...
   }
 
+  @Test public void installEmptyCategory() {
+    final EmojiProvider emptyProvider = new EmojiProvider() {
+      @NonNull @Override public EmojiCategory[] getCategories() {
+        return new EmojiCategory[0];
+      }
+    };
+
+    assertThatThrownBy(new ThrowableAssert.ThrowingCallable() {
```
vanniktech: uh nice. TIL 👍

emoji/src/test/java/com/vanniktech/emoji/EmojiManagerTest.java
```diff
@@ -49,10 +47,26 @@
...
+      @Override
+      public void call() throws Throwable {
+        EmojiManager.install(emptyProvider);
+      }
+    }).isInstanceOf(IllegalArgumentException.class);
```
vanniktech: Can you also verify the message here?

emoji/src/main/res/values/library_emoji_strings.xml
rubengees: Lint complained about unused resources here, failing the build.

emoji/src/test/java/com/vanniktech/emoji/EmojiManagerTest.java
vanniktech: also squeeze in another test case where you pass `emptyList()` as the providers

emoji/src/main/java/com/vanniktech/emoji/RecentEmojiManager.java
```diff
@@ -133,7 +133,7 @@ int size() {
...
+    @Override @NonNull public Iterator<Data> iterator() {
```
rubengees: Also, lint was complaining about this one.

emoji/src/test/java/com/vanniktech/emoji/EmojiManagerTest.java
```diff
@@ -68,4 +82,10 @@
 
     assertThat(EmojiManager.getInstance().findEmoji("")).isNull();
   }
+
+  @Test public void findAllEmojisEmpty() {
```
vanniktech: this only tests the case of `findAllEmojis` when it's empty. It should also test the actual logic if it's not empty

emoji/src/main/res/values/library_emoji_strings.xml
vanniktech: I think it's a lint bug.

`emoji/lint.xml` contains this:

```xml
<issue id="UnusedResources">
    <ignore regexp="library_emoji_strings.xml"/>
</issue>
```

https://issuetracker.google.com/issues/37981147

rubengees: Should we keep this ignore in for the moment, until fixed by the Google team?

emoji/src/main/java/com/vanniktech/emoji/EmojiHandler.java
rubengees: Not necessarily needed anymore, as we access the fields of `EmojiRange` directly now.

emoji/src/main/java/com/vanniktech/emoji/EmojiHandler.java
rubengees: It hurts my eyes :cry:. Required?

emoji/src/test/java/com/vanniktech/emoji/EmojiManagerTest.java
rubengees: I guess you meant an empty `EmojiCategory`? If not, could you go in detail what you mean?

emoji/src/test/java/com/vanniktech/emoji/EmojiManagerTest.java
rubengees: This is already [entirely covered] by the tests in the `EmojiHandlerTest`. Should i test it nevertheless here?

rubengees:
Most of it done (I have added comments where not).
I also update everything to the respective latest stable versions, while i was at it. If not desired, let me know.

emoji/src/test/java/com/vanniktech/emoji/EmojiManagerTest.java
```diff
@@ -49,10 +47,61 @@
     assertThat(EmojiManager.getInstance().getCategories()).isNotEmpty();
   }
 
+  @Test
+  public void noProviderInstalled() {
```
rubengees: This test is currently failing because `EmojiManager` is a singleton and there is a `EmojiProvider` from a previous test installed.
This led me to the thought of adding a `EmojiManager#destroy` method which would remove the current `EmojiProvider` and clean up all internal data-structures.
This would allow apps to free the memory when they know that the `EmojiManager` is not needed anymore (The user exits the screen with the emoji input for example).

What do you think of this?

emoji/src/main/res/values/library_emoji_strings.xml
vanniktech: just removei t from `lint.xml` if it works in the xml

emoji/src/main/java/com/vanniktech/emoji/EmojiHandler.java
vanniktech: Yes I also realized this then. Even better 👍

emoji/src/main/java/com/vanniktech/emoji/EmojiHandler.java
vanniktech: Nah then leave it as is. Or if you already changed it, leave it. Up to you. I've got a big monitor and prefer having it in one but having two lines is also ok

emoji/src/test/java/com/vanniktech/emoji/EmojiManagerTest.java
```diff
@@ -49,10 +47,61 @@
...
+      public void call() throws Throwable {
+        EmojiManager.getInstance().findEmoji("test");
+      }
+    }).isInstanceOf(IllegalStateException.class).hasMessage("Please install an EmojiProvider through the EmojiManager.install() method first.");
+  }
+
+  @Test public void installEmptyProvider() {
+    final EmojiProvider emptyProvider = new EmojiProvider() {
+      @NonNull @Override public EmojiCategory[] getCategories() {
+        return new EmojiCategory[0];
```
vanniktech: Also a test here for having a category but the category not having any Emojis.

emoji/src/test/java/com/vanniktech/emoji/EmojiManagerTest.java
vanniktech: Just one simple one so that it does not look like it's not covered.

build.gradle
```diff
@@ -70,16 +70,16 @@ ext {
...
+            assertJ           : 'org.assertj:assertj-core:3.8.0'
     ]
 
     javaVersion = JavaVersion.VERSION_1_7
 }
 
 task wrapper(type: Wrapper) {
-    gradleVersion = '3.3'
+    gradleVersion = '3.5'
     distributionUrl = "https://services.gradle.org/distributions/gradle-$gradleVersion-all.zip"
```
vanniktech: when we are anyways changing everything this line can be changed to `distributionType = Wrapper.DistributionType.ALL`

build.gradle
```diff
@@ -7,7 +7,7 @@ codeQualityTools {
     toolVersion = '7.7'
```
vanniktech: 7.8 🤓

emoji/src/main/java/com/vanniktech/emoji/EmojiManager.java
```diff
@@ -13,9 +20,18 @@
  */
 public final class EmojiManager {
   private static final EmojiManager INSTANCE = new EmojiManager();
+  private static final int GUESSED_UNICODE_AMOUNT = 3000;
+  private static final int GUESSED_TOTAL_PATTERN_LENGTH = GUESSED_UNICODE_AMOUNT * 4;
 
-  private final EmojiTree emojiTree = new EmojiTree();
+  private static final Comparator<String> STRING_LENGTH_COMPARATOR = new Comparator<String>() {
+    @Override public int compare(final String first, final String second) {
+      return second.length() - first.length();
```
vanniktech: This can cause issues. Have a look at [this](https://stackoverflow.com/a/2728858/1979703). It's an extreme edge case though, although we should cover it.

emoji/src/main/java/com/vanniktech/emoji/EmojiManager.java
```diff
@@ -34,32 +50,94 @@ static EmojiManager getInstance() {
...
       //noinspection ForLoopReplaceableByForEach
       for (int j = 0; j < emojis.length; j++) {
-        INSTANCE.emojiTree.add(emojis[j]);
+        final Emoji emoji = emojis[j];
+        final String unicode = emoji.getUnicode();
+        final List<Emoji> variants = emoji.getVariants();
+
+        INSTANCE.emojiMap.put(unicode, emoji);
+        unicodesForPattern.add(unicode);
+
```
vanniktech: `//noinspection ForLoopReplaceableByForEach` so that intellij won't complain

emoji/src/test/java/com/vanniktech/emoji/EmojiManagerTest.java
```diff
@@ -49,10 +47,61 @@
     assertThat(EmojiManager.getInstance().getCategories()).isNotEmpty();
   }
 
+  @Test
```
vanniktech: same line

emoji/src/test/java/com/vanniktech/emoji/EmojiManagerTest.java
vanniktech: Let's make it for now package private so it can't be used. I anyways want the install() method to provide some type on which you can do a bunch of strings (like the string queries that I proposed) and this method might get added too.

emoji/src/test/java/com/vanniktech/emoji/EmojiManagerTest.java
vanniktech: See above.

emoji/src/main/java/com/vanniktech/emoji/EmojiManager.java
rubengees: Oh right! Never thought about it that way.

rubengees:
All done :+1:

vanniktech:
I just wanted to update the branch too. You were a bit faster though.

Once it's done, I'll merge it. Many thanks 👍

rubengees:
Great! What would be missing for a `0.5` release?

vanniktech:
Not much. Just haven't gotten around to cutting a release yet plus my release computer isn't set up right now. Although I'm planning to release 0.5.0 next week.

PR #132, opened 2017-06-01
New emojis

rubengees:
- Google has updated their emoji set and this PR updates that.
- EmojiOne also seem to have some changes, although not directly visible.
- I adjusted the script to work with the lastest files (Why do these change so often?) and added logs for emojis that were not found (The `emoji.json` that we are using does not have entries for the latest additions yet. I also did not found any other file with these in).


vanniktech:
Dunno why they change so often. Thanks for updating it 👍

PR #143, opened 2017-06-22
Don't generate BuildConfig file for release builds.

vanniktech:


rubengees:
Just interested: Why is it better to not generate one?

vanniktech:
It's not needed. So we can skip the generation plus when you have a lot of libraries you don't end up with 100 entries for different BuildConfig files.

rubengees:
I see, thanks for the explanation!

PR #149, opened 2017-06-27
Latest skin tone is remembered as new default

RicoChr:
A problem i noticed is that we now have a slight flicker when the gridview is updated due to the change of the default icon (the base emoji is never changed of course, just which one is displayed!)
Apart from that, this should be fully functional.

### DIFF
--- a/emoji/src/main/java/com/vanniktech/emoji/EmojiPagerAdapter.java
+++ b/emoji/src/main/java/com/vanniktech/emoji/EmojiPagerAdapter.java
@@ -30,18 +30,14 @@ final class EmojiPagerAdapter extends PagerAdapter {
 
   @Override public Object instantiateItem(final ViewGroup pager, final int position) {
     final View newView;
-
     if (position == 0) {
       newView = new RecentEmojiGridView(pager.getContext()).init(listener, longListener, recentEmoji);
-
       recentEmojiGridView = (RecentEmojiGridView) newView;
     } else {
       newView = new EmojiGridView(pager.getContext()).init(listener, longListener, EmojiManager.getInstance().getCategories()[position - 1]);
       gridViewsList.add((EmojiGridView) newView);
     }
-
     pager.addView(newView);
-
     return newView;
   }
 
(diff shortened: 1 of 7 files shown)


RicoChr:
these code-style checks are ridiculous...

vanniktech:
Kinda. If you want I can take care of them for you.

RicoChr:
That would be nice.
Have you integrated these checks into your development environment? If so, i would like to know how.

vanniktech:
I've got a [plugin](https://github.com/vanniktech/gradle-code-quality-tools-plugin) which lets you apply `PMD`, `Findbugs` and `Checkstyle` to any java | android project. And the configuration itself is under `code_quality_tools`. I just execute them via the terminal to see what's failing `./gradlew checkstyle findbugs pmd`.

emoji/src/main/java/com/vanniktech/emoji/EmojiPagerAdapter.java
```diff
@@ -27,23 +30,26 @@
...
       newView = new EmojiGridView(pager.getContext()).init(listener, longListener, EmojiManager.getInstance().getCategories()[position - 1]);
+      gridViewsList.add((EmojiGridView) newView);
     }
-
     pager.addView(newView);
-
     return newView;
   }
 
+  public void notifyPagesChanged(){
```
vanniktech: this is probably causing the lag. Also it's not optimal. How about just changing the page that needs to be changed? (it's the one you're in and the recent one.

RicoChr: I couldn't find yet find out, how to get the current page so i just updated them all

vanniktech: There are methods inside the ViewPager for this.

RicoChr: I can get the index of the page but i have no clue on how to get the corresponding gridView. I think my gridViewsList is not in order...

vanniktech:
I'll close this in favor of #157. I've taken your first commit and worked a bit around it so that we only update the ImageView and not even the entire GridView.

Thanks for your initial work.

RicoChr:
You are welcome, this is still a very much a Challenge for me and i enjoy it :)

PR #157, opened 2017-06-28
Reflect Emoji Skin Tone automatically in Emoji List.

### LINKED ISSUE #101: Reflect Emoji Skin Tone automatically in Emoji List
When changing the skin tone of an Emoji in a Category the Emoji Icon in the list should be updated. 

1. Look for 👊
2. Long press on 👊 
3. Change in Dialog to 👊🏻
4. In the list 👊🏻 should be displayed now instead of 👊

If we do this we should no longer then display 👊 & 👊🏻 in the recent Emojis. Instead just display the one that was used last.

vanniktech:
Partially taken from #149. Some things are still left to do:

- [x] Still a little lag between showing the new Emoji and the old one
- [x] Some skin modifiers are not working properly (for some reason some of the guys that are only 4 characters long e.g. 270a)
- [x] Tests

Fixes #101

### DIFF
--- a/emoji/src/main/java/com/vanniktech/emoji/EmojiImageView.java
+++ b/emoji/src/main/java/com/vanniktech/emoji/EmojiImageView.java
@@ -5,10 +5,14 @@ import android.graphics.Canvas;
 import android.graphics.Paint;
 import android.graphics.Path;
 import android.graphics.Point;
+import android.support.annotation.Nullable;
 import android.support.annotation.RestrictTo;
 import android.support.v4.content.ContextCompat;
 import android.support.v7.widget.AppCompatImageView;
 import android.util.AttributeSet;
+import android.view.View;
+import com.vanniktech.emoji.emoji.Emoji;
+import com.vanniktech.emoji.listeners.OnEmojiClickedListener;
 
 import static android.support.annotation.RestrictTo.Scope.LIBRARY;
 
@@ -24,6 +28,8 @@ import static android.support.annotation.RestrictTo.Scope.LIBRARY;
   private final Point variantIndicatorBottomLeft = new Point();
 
   private boolean hasVariants;
+
+  private RecentEmojiVariantManager.Disposable disposable;
 
   public EmojiImageView(final Context context, final AttributeSet attrs) {
     super(context, attrs);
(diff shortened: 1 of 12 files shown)


emoji/src/main/java/com/vanniktech/emoji/EmojiImageView.java
```diff
@@ -66,9 +71,65 @@ public EmojiImageView(final Context context, final AttributeSet attrs) {
...
+          }
+
+          return false;
+        }
+      });
+    } else {
+      setOnLongClickListener(null);
+    }
+
+    ImageLoadingTask task = (ImageLoadingTask) getTag();
```
rubengees: Maybe avoid saving the `task` in the tag altogether and just call `dispose` on the `Disposable` class member?

emoji/src/main/java/com/vanniktech/emoji/RecentVariantManager.java
```diff
@@ -0,0 +1,145 @@
...
+import android.content.SharedPreferences;
+import android.support.annotation.NonNull;
+import com.vanniktech.emoji.emoji.Emoji;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.StringTokenizer;
+
+final class RecentVariantManager { // TODO: test it
```
rubengees: Make this an `interface` like the `RecentEmojiManger` so consumers could choose to implement persistance themselfs?

emoji/src/main/java/com/vanniktech/emoji/RecentVariantManager.java
```diff
@@ -0,0 +1,145 @@
...
+    //}
+
+    // TODO: sometimes those 4 characters are not working (e.g. 270a)
+
+    for (final Emoji recentEmoji : recentVariantsList) {
+      if (recentEmoji.getBase().equals(newVariantBase)) {
+        if (recentEmoji.equals(newVariant)) { // Same skin-tone was used.
+          return;
+        } else {
+          recentVariantsList.remove(recentEmoji);
```
rubengees: This is potentially slower than possible.

`remove` needs to find the `recentEmoji` again, by iteration the list, even though we have already found it.
When using `fori` in which we memorize the index, we can avoid that.

rubengees:
The lag does not look too bad, WhatsApp also has it if you look closely.

emoji/src/main/java/com/vanniktech/emoji/EmojiImageView.java
vanniktech: I think the Task is only necessary when at once trying to display all the Emojis. I don't think we need it to just switch the variant (We know for sure that you can only do this one by one). I think I'll just end up calling `setImageResource()` and also avoid the `setDrawable(null)` call which should get rid of the flickering.

emoji/src/main/java/com/vanniktech/emoji/RecentVariantManager.java
vanniktech: Thought about it and I don't think that you want to do this. If there's high demand for it we can still expose it. I prefer to keep as many internals internals.

emoji/src/main/java/com/vanniktech/emoji/RecentVariantManager.java
vanniktech: Yes definitely, it's also the implementation from #149 and I'm still going to optimize it but first I want to add some tests as the current implementation does seem to work (except for the edge case outlined above)

vanniktech:
Yup also saw the lag there, I hope I can get rid of it, otherwise the functionality gain (not having to always re-select your favorite skin tone) outweighs the small lag that you can see.

PR #158, opened 2017-06-29
Create Height Offset

searchy2:
*Create Emoji Builder method to add height offset to EmojiPopup.

### DIFF
--- a/emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
+++ b/emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
@@ -25,248 +25,257 @@ import com.vanniktech.emoji.listeners.OnSoftKeyboardOpenListener;
 import static com.vanniktech.emoji.Utils.checkNotNull;
 
 public final class EmojiPopup {
-  static final int MIN_KEYBOARD_HEIGHT = 100;
+    static final int MIN_KEYBOARD_HEIGHT = 100;
 
-  final View rootView;
-  final Activity context;
+    final View rootView;
+    final Activity context;
 
-  @NonNull final RecentEmoji recentEmoji;
-  @NonNull final EmojiVariantPopup variantPopup;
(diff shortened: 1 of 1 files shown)


searchy2:
Apologies for the reformatting.

searchy2:
Is this the appropriate style? https://github.com/square/java-code-styles
I am having some trouble applying it. Importing into Android Studio does not seem to work.

vanniktech:
Yup it is.

searchy2:
`Intellij IDEA code style XML import failed with error message: Can't read fromSquareAndroid.xml, Error on line 42: Attribute name "data-pjax-transient" associated with an element type "meta" must be followed by the ' = ' character.`

This is the error I'm getting when trying to import. Can you fix the indentations instead?

vanniktech:
I'm not even sure I want this property as I said.

searchy2:
It would be useful in cases like mine where the only way to get proper layouting is to add a manual offset. Otherwise, there is no easy way to fix the gap.

vanniktech:
My problem is that this hides the underlying problem (Dialog height calculated wrong in the first place). Can you check whether if you enable immersive mode in the sample app you get the same behavior / bug?

searchy2:
The problem is that there is no problem haha. In immersive mode, the sample app calculates the height of the keyboard correctly. Your code works perfectly! :+1: 

I have attempted to isolate the cause of the issue in my app but have been unable to do so. Again, there is nothing wrong with your library. In my app, I find the offset is the same value (after dp->px) on every device so an offset would fix the issue for me.

vanniktech:
I'll close this then since it's a problem on your end and I don't want to burden the API of this library with this extra method that others may get confused about. Also you should try to fix the problem on your side rather than proposing a solution to every library that will be hit by this.

Another idea might be that `fitSystemWindows` is responsible for that effect.

PR #165, opened 2017-07-01
Refactor the VariantEmojiManager

### LINKED ISSUE #164: RecentEmojiManager leaks
- Version of the library: Latest Snapshot (cf6b2169297010bcb3552f0f7bc08b28fabe696c)
- Affected devices: All
- Affected versions: Tested on `7.1.2`

This is the leak trace from `leakcanary`:

```
* GC ROOT static com.vanniktech.emoji.RecentEmojiVariantManager.INSTANCE
* references com.vanniktech.emoji.RecentEmojiVariantManager.variantListeners
* references java.util.HashMap.table
* references array java.util.HashMap$HashMapEntry[].[1]
* references java.util.HashMap$HashMapEntry.value
* references com.vanniktech.emoji.EmojiImageView$1.val$clickListener (anonymous
  implementation of com.vanniktech.emoji.RecentEmojiVariantManager$EmojiVariantListener)
* references com.vanniktech.emoji.EmojiPopup$3.this$0 (anonymous implementation
  of com.vanniktech.emoji.listeners.OnEmojiClickedListener)
* references com.vanniktech.emoji.EmojiPopup.onEmojiPopupDismissListener
* <my code>
```

This seems to happen due to the variant listeners which never get cleaned up.

rubengees:
This should avoids leaks and improve the architecture. The behaviour is not changed. Would fix #164.

### Code changes to achieve this

- New `interface` `VariantEmoji` similar to `RecentEmoji`. This is exposed through the `builder` and can be implemented by users for saving somewhere else than the `SharedPreferences`. `VariantEmojiManager` is the default implementation of that, similar to the `RecentEmojiManager`.
- `VariantEmojiManager` is not a singleton anymore. This leads to passing the object around anywhere, but thats better than a singleton.
- The `EmojiImageView` now gets updated on click, instead of implicitely by a listener. This is better as it leads to lower coupling.
- Various renamings and smaller clean ups. I thought this might be a good idea; We already have breaking changes, so now is the time to clean up :grin:.
  - The `EmojiOnClickListener` has been extended to also have the `EmojiImageView` as a parameter. I needed that for updating the view properly, but it's also consistent with Android's listeners, which also have the `View` as parameter.


app/src/main/java/com/vanniktech/emoji/sample/MainActivity.java
```diff
@@ -128,12 +129,13 @@
...
         .setOnEmojiBackspaceClickListener(new OnEmojiBackspaceClickListener() {
-          @Override public void onEmojiBackspaceClicked(final View v) {
+          @Override public void onEmojiBackspaceClick(final View v) {
             Log.d(TAG, "Clicked on Backspace");
           }
         })
-        .setOnEmojiClickedListener(new OnEmojiClickedListener() {
-          @Override public void onEmojiClicked(@NonNull final Emoji emoji) {
+        .setOnEmojiClickListener(new OnEmojiClickListener() {
+          @Override
```
vanniktech: nit: can you move it onto the same line as below?

app/src/main/java/com/vanniktech/emoji/sample/MainDialog.java
```diff
@@ -102,12 +103,13 @@ private View buildView() {
...
             .setOnEmojiBackspaceClickListener(new OnEmojiBackspaceClickListener() {
-              @Override public void onEmojiBackspaceClicked(final View v) {
+              @Override public void onEmojiBackspaceClick(final View v) {
                 Log.d(TAG, "Clicked on Backspace");
               }
             })
-            .setOnEmojiClickedListener(new OnEmojiClickedListener() {
-              @Override public void onEmojiClicked(@NonNull final Emoji emoji) {
+            .setOnEmojiClickListener(new OnEmojiClickListener() {
+              @Override
```
vanniktech: same

emoji/src/main/java/com/vanniktech/emoji/EmojiImageView.java
```diff
@@ -72,78 +79,55 @@ public EmojiImageView(final Context context, final AttributeSet attrs) {
...
-          @Override public void onChanged(final Emoji newEmoji) {
-            setImageResource(newEmoji.getResource()); // We don't need to load it in an async manner since it's only one Image and we know that for a fact.
-
-            setOnClickListener(new EmojiOnClickListener(clickListener, emoji));
-          }
-        });
-  }
-
-  @Override protected void onDetachedFromWindow() {
+  @Override
```
vanniktech: nit: same line as below

emoji/src/main/java/com/vanniktech/emoji/EmojiArrayAdapter.java
```diff
@@ -8,21 +8,25 @@
...
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 
 import static com.vanniktech.emoji.Utils.checkNotNull;
 
 final class EmojiArrayAdapter extends ArrayAdapter<Emoji> {
-  @Nullable private final OnEmojiClickedListener listener;
-  @Nullable private final OnEmojiLongClickedListener longListener;
+  @Nullable private final VariantEmoji variantManager;
```
vanniktech: Should be @NonNull right since you create it initially in the Popup

emoji/src/main/java/com/vanniktech/emoji/EmojiVariantPopup.java
```diff
@@ -24,18 +24,21 @@
...
+  EmojiVariantPopup(@NonNull final View rootView, @Nullable final OnEmojiClickListener listener) {
     this.rootView = rootView;
     this.listener = listener;
   }
 
-  void show(@NonNull final View clickedImage, @NonNull final Emoji emoji) {
+  void show(@NonNull final EmojiImageView clickedImage, @NonNull final Emoji emoji) {
     dismiss();
 
+    this.rootImageView = clickedImage;
```
vanniktech: nit: remove `this.`

emoji/src/main/java/com/vanniktech/emoji/EmojiVariantPopup.java
```diff
@@ -57,6 +60,8 @@ void show(@NonNull final View clickedImage, @NonNull final Emoji emoji) {
   }
 
   void dismiss() {
+    this.rootImageView = null;
```
vanniktech: nit: remove `this.`

emoji/src/main/java/com/vanniktech/emoji/EmojiArrayAdapter.java
rubengees: No, the `EmojiArrayAdapter` of the recent tab gets `null` passed in for the `variantManager`, because we do not want to show what the `variantManager` returns, but that what the `recentManager` returns.

Yes, this is a bit hacky, we could also have a dedicated `EmojiArrayAdapter` for that, but I thought that would be a bit of overkill.

rubengees:
Already fixed, working on some checkstyle issues then I'll push that change also.

rubengees:
Done :+1:

emoji/src/main/java/com/vanniktech/emoji/EmojiImageView.java
```diff
@@ -122,11 +122,26 @@ public boolean onLongClick(final View view) {
...
 
-  public void setOnEmojiClickListener(@Nullable final OnEmojiClickListener listener) {
+  /**
+   * Updates the emoji image directly. This should be called only for updating the variant
+   * displayed (of the same base emoji), since it does not run asynchronously and does not update
+   * the internal listeners.
+   *
+   * @param emoji The new emoji variant to show.
+   */
+  void updateEmoji(@NonNull final Emoji emoji) {
```
vanniktech: should this be public? otherwise you can't access it when you want to do your custom implementation

emoji/src/main/java/com/vanniktech/emoji/EmojiImageView.java
```diff
@@ -88,7 +88,7 @@ public EmojiImageView(final Context context, final AttributeSet attrs) {
     }
   }
 
-  public void setEmoji(@NonNull final Emoji emoji) {
+  void setEmoji(@NonNull final Emoji emoji) {
```
vanniktech: should this be public? otherwise you can't access it when you want to do your custom implementation

rubengees: This is not supposed to be overwritten, the class itself is only `public`, because of the package structure (it's used in some listeners, which reside in a different package). For that reason it has the `@RestrictTo(LIBRARY)` annotation.

rubengees:
`EmojiImageView` is `final` or do you mean another class?

PR #166, opened 2017-07-01
Update generator for latest changes

rubengees:
This updates the `generator` for the latest changes, which are:

- `NodeJs` 8.
- Adjustments for the latest emoji metadata files.
- Latest library versions.

I have also included a flag for a larger heap. The default does not seem to be enough with the latest additions.

There are quite some new emojis, but I haven't included them, because they are [not ordered correctly](https://github.com/emojione/emojione/issues/523) yet.

### DIFF
--- a/UPDATING.md
+++ b/UPDATING.md
@@ -14,7 +14,7 @@ The script does five things:
 
 ## Prerequisites
 
-[NodeJS 7](https://nodejs.org)<br>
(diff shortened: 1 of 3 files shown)


rubengees:
Yes, I have seen the ios ones on the unicode website in another document already, but they are not included in the file we are parsing yet  :cry:. I'll check from time to time.

PR #167, opened 2017-07-01
Update Google Emojis.

vanniktech:
Just because emojione does not seem to be working we can still update the google ones.


rubengees:
The problem is the messed up order. The new faces are at the bottom whereas I would expect them to be near the other faces. This is true for all emoji providers, since the same (not yet correct) file is used for them (that one from the `emojione` repository)

I would recommend to wait until they fix that.

vanniktech:
Oh okay. Now I got it. I also was a bit confused since the java diff was the same. Okay thanks then I'll close this and wait.

PR #188, opened 2017-08-07
Dependency updates

rubengees:
As the title says.


app/src/main/java/com/vanniktech/emoji/sample/MainActivity.java
```diff
@@ -30,6 +30,7 @@
 import com.vanniktech.emoji.one.EmojiOneProvider;
 import com.vanniktech.emoji.twitter.TwitterEmojiProvider;
 
+@SuppressWarnings("CPD-START") // We don't care about duplicate code in the sample.
```
rubengees: Found here: https://stackoverflow.com/a/18625921/4279995
Not optimal, but I think there is no better way...

build.gradle
```diff
@@ -17,19 +17,21 @@ buildscript {
...
             ],
         ],
-        'fastLaneScreenGrab': "tools.fastlane:screengrab:1.0.3",
+        'fastLaneScreenGrab': "tools.fastlane:screengrab:1.1.0",
     ]
 
     repositories {
         jcenter()
         google()
+        maven { url "https://plugins.gradle.org/m2/" }
```
rubengees: We need this for the `gradle-errorprone-plugin`.

emoji/src/test/java/com/vanniktech/emoji/EmojiManagerTest.java
```diff
@@ -189,16 +189,6 @@ public void noProviderInstalled() {
     assertThat(text.getSpans(0, text.length(), EmojiSpan.class)).hasSize(1);
   }
 
-  @Test public void takeLongest() {
```
rubengees: This was a duplicate of the `halfPath` test.

build.gradle
vanniktech: We don't. The code quailty plugin takes care of that.

build.gradle
```diff
@@ -4,7 +4,7 @@ apply plugin: 'com.vanniktech.android.junit.jacoco'
 
 buildscript {
     ext.versions = [
-        testRules: '0.5',
+        testRules: '1.0.0',
```
vanniktech: i haven't had luck with 1.0.0 and espresso 3.0.0. Did you test locally whether the screenshots are being taken or not. You can do so using `gw assembleDebug assembleAndroidTest && fastlane screengrab`

build.gradle
```diff
@@ -17,19 +17,21 @@ buildscript {
...
+        maven { url "https://plugins.gradle.org/m2/" }
     }
     dependencies {
         classpath 'com.vanniktech:gradle-code-quality-tools-plugin:0.7.0'
-        classpath 'com.vanniktech:gradle-android-junit-jacoco-plugin:0.6.0'
+        classpath 'com.vanniktech:gradle-android-junit-jacoco-plugin:0.7.0'
 
         classpath 'com.android.tools.build:gradle:2.3.3'
         classpath 'com.github.ben-manes:gradle-versions-plugin:0.15.0'
+        classpath 'net.ltgt.gradle:gradle-errorprone-plugin:0.0.11'
```
vanniktech: not needed. The code quailty plugin takes care of that.

build.gradle
rubengees: Oops, I'll revert it.

build.gradle
rubengees: You are right, at first it did not work. Managed to fix with my latest commit though.

build.gradle
```diff
@@ -4,7 +4,8 @@ apply plugin: 'com.vanniktech.android.junit.jacoco'
 
 buildscript {
     ext.versions = [
-        testRules: '0.5',
+        espresso: '3.0.0',
```
rubengees: Concerning screengrab: Something set a different version, so I redefinied it here. Why does that happen?

build.gradle
rubengees: Fixed.

build.gradle
vanniktech: Oh it was simply missing here in the first place. Now it should be all good.

PR #196, opened 2017-08-26
Emoji Support Library integration

stefanhaustein:
Quite minimal

Potential followups:
- Emoji API and class hierarchy cleanup
- Generator support
- Route span generation through the provider to reduce wrapping levels
- readme.md


app/src/main/java/com/vanniktech/emoji/sample/EmojiApplication.java
```diff
@@ -17,6 +19,10 @@
       return;
     }
 
+    final EmojiCompat.Config config = new BundledEmojiCompatConfig(this);
```
vanniktech: can we move this into the Activity where we'll also install the provider?

app/src/main/res/layout/activity_main.xml
```diff
@@ -39,9 +39,8 @@
             android:layout_width="0dp"
             android:layout_height="wrap_content"
             android:layout_weight="1"
-            android:imeOptions="actionSend"
-            android:inputType="textCapSentences|textMultiLine"
-            android:maxLines="3"/>
```
vanniktech: can you restore the max lines?

emoji-google-compat/build.gradle
```diff
@@ -0,0 +1,28 @@
+apply plugin: 'com.android.library'
+
+afterEvaluate {
+  generateReleaseBuildConfig.enabled = false
+}
+
+android {
+    compileSdkVersion rootProject.ext.compileSdkVersion as int
```
vanniktech: can you do 2 tabs instead of 4?

emoji-google-compat/src/main/java/com/vanniktech/emoji/googlecompat/GoogleCompatEmoji.java
```diff
@@ -0,0 +1,30 @@
+package com.vanniktech.emoji.googlecompat;
+
+import android.content.Context;
+import android.graphics.drawable.Drawable;
+import android.support.annotation.NonNull;
+import com.vanniktech.emoji.emoji.Emoji;
+
+public class GoogleCompatEmoji extends Emoji {
```
vanniktech: let's make it final so others can't extend it

emoji-google-compat/src/main/java/com/vanniktech/emoji/googlecompat/GoogleCompatEmoji.java
```diff
@@ -0,0 +1,30 @@
...
+  public GoogleCompatEmoji(final int[] ints, final Emoji... emojis) {
+    super(ints, -1, emojis);
+  }
+
+  public GoogleCompatEmoji(final int codePoint, final Emoji... emojis) {
+    super(codePoint, -1, emojis);
+  }
+
+  @NonNull
+  @Override
```
vanniktech: nit: sameline as the method

emoji-google-compat/src/main/java/com/vanniktech/emoji/googlecompat/GoogleCompatEmoji.java
```diff
@@ -0,0 +1,30 @@
...
+
+  public GoogleCompatEmoji(final int[] ints, final Emoji... emojis) {
+    super(ints, -1, emojis);
+  }
+
+  public GoogleCompatEmoji(final int codePoint, final Emoji... emojis) {
+    super(codePoint, -1, emojis);
+  }
+
+  @NonNull
```
vanniktech: nit: sameline as the method

emoji-google-compat/src/main/java/com/vanniktech/emoji/googlecompat/GoogleCompatEmoji.java
```diff
@@ -0,0 +1,30 @@
...
+  @Override
+  public Drawable getDrawable(final Context context) {
```
vanniktech: nit: `@NonNull final Context`

emoji-google-compat/src/main/java/com/vanniktech/emoji/googlecompat/GoogleCompatEmojiDrawable.java
```diff
@@ -0,0 +1,64 @@
...
+  @Override
+  public void draw(final Canvas canvas) {
+    final Rect bounds = getBounds();
+    textPaint.setTextSize(bounds.height() * TEXT_SIZE_FACTOR);
+    final int y = Math.round(bounds.bottom - bounds.height() * BASELINE_OFFSET_FACTOR);
+
+    if (emojiSpan == null) {
+      canvas.drawText(processed, 0, processed.length(), bounds.left, y, textPaint);
+    } else {
+      emojiSpan.draw(canvas, processed, 0, processed.length(), bounds.left /* x */, bounds.top /* top */, y /* y */, bounds.bottom /* bottom */, textPaint);
```
vanniktech: can you remove the comments regarding the parameter names?

emoji-google-compat/src/main/java/com/vanniktech/emoji/googlecompat/GoogleCompatEmojiDrawable.java
```diff
@@ -0,0 +1,64 @@
...
+      final Object[] spans = ((Spanned) processed).getSpans(0, unicode.length(), EmojiSpan.class);
+      if (spans.length > 0) {
+        emojiSpan = (EmojiSpan) spans[0];
+      }
+    }
+    textPaint.setStyle(Paint.Style.FILL);
+    textPaint.setColor(0x0ffffffff);
+  }
+
+  @Override
```
vanniktech: nit: same line

emoji-google-compat/src/main/java/com/vanniktech/emoji/googlecompat/GoogleCompatEmojiDrawable.java
```diff
@@ -0,0 +1,64 @@
...
+import android.graphics.drawable.Drawable;
+import android.support.text.emoji.EmojiCompat;
+import android.support.text.emoji.EmojiSpan;
+import android.text.Spanned;
+import android.text.TextPaint;
+
+/**
+ * An emoji drawable backed by a span generated by the Google emoji support library.
+ */
+public class GoogleCompatEmojiDrawable extends Drawable {
```
vanniktech: final & package private so others can't use it and we won't have to support it in the public API

emoji-google-compat/src/main/java/com/vanniktech/emoji/googlecompat/GoogleCompatEmojiDrawable.java
```diff
@@ -0,0 +1,64 @@
...
+    final int y = Math.round(bounds.bottom - bounds.height() * BASELINE_OFFSET_FACTOR);
+
+    if (emojiSpan == null) {
+      canvas.drawText(processed, 0, processed.length(), bounds.left, y, textPaint);
+    } else {
+      emojiSpan.draw(canvas, processed, 0, processed.length(), bounds.left /* x */, bounds.top /* top */, y /* y */, bounds.bottom /* bottom */, textPaint);
+    }
+  }
+
+  @Override
```
vanniktech: nit: same line as method function

emoji-google-compat/src/main/java/com/vanniktech/emoji/googlecompat/GoogleCompatEmojiDrawable.java
```diff
@@ -0,0 +1,64 @@
...
+  public void setAlpha(final int alpha) {
+    textPaint.setAlpha(alpha);
+  }
+
+  @Override
```
vanniktech: nit: same line as method function

emoji-google-compat/src/main/java/com/vanniktech/emoji/googlecompat/GoogleCompatEmojiDrawable.java
```diff
@@ -0,0 +1,64 @@
...
+  public void setColorFilter(final ColorFilter colorFilter) {
+    textPaint.setColorFilter(colorFilter);
+  }
+
+  @Override
```
vanniktech: nit: same line as method function

stefanhaustein:
Done. Have created the categories manually by changing the generated google categories. Wanted to look into how sprite sheet metadata is handled for EmonjiOne etc. before changing the generator (and don't have nodejs installed on the laptop).

rubengees:
@stefanhaustein This is great :+1:! How do the categories differ from the normal Google ones?
@vanniktech I could follow up with the generator adjustments once EmojiOne pushes fixes for their `emoji.json` which we are using for generation (There is a problem with the ordering currently).

vanniktech:
The categories should not differ at all it's just that for this artifact the google support library will take care of the images and provide them lazily via fonts so they won't blow up the apk.

Adjusting the generator sounds perfect! Yup watched that issue regarding the ordering and there's no progress yet.

stefanhaustein:
The categories should be identical to the regular Google categories.

The support library can be used either with bundled emoji (as currently in the demo app), or the initialization can download the font as needed (= not on Oreo). Both setup methods have different dependencies, which makes it tricky to include the library setup, see https://developer.android.com/guide/topics/ui/look-and-feel/emoji-compat.html

rubengees:
@stefanhaustein Thanks for the clarification.

I think we could also implement the lazy approach of the `emoji-compat` lib in the long run and move away from our strict initialization (With warnings to avoid developers not calling `init`!).

`EmojiTextView` and the other views would then simply leave the `String` untouched (or cut out emojis to avoid the missing font characters), if uninitialized. This would also not change the behaviour of current `Providers` because initialization is instantious for them.

I think this would require separate modules for `emoji-compat` and `emoji-compat-bundled` though.

Thoughts?

stefanhaustein:
I think it would be fine to leave the support lib setup to the user. The download case requires some download key metadata in the manifest anyway IIRC, and users might also want to use the support lib elsewhere, which might make full wrapping awkward. 

One option to make the dependency more explicit might be to hand in the emoji support lib config object to the GoogleCompatEmojiProvider constructor (and just make sure it's not null).

We could still add emoji-compat-bundled on demand later if the extra setup really turns out to be a problem for users.

To me, the most important next step seems to be moving span generation to the provider, so we can avoid the span -> drawable -> span indirection in the support library case.

rubengees:
@stefanhaustein Ah sorry, I misunderstood how the `emoji-support` lib works, we actually already have the behaviour I was thinking of.

Concerning the dependency explicity: I think there is room for improvements (The setup could be plug-and-play with optional configuration if wanted/needed), but you are right, other things like lib size and the span code are more important for now.

Involving this PR, I have found an issue while testing it locally. Consider these screenshots:

| `emoji-compat` | Existing `google` |
| --- | --- |
| screenshot_20170830-000128 | screenshot_20170830-000138 |

The new `Provider` seems to be missing some emojis. Is this intentional (Maybe these are unsupported by `emoji-compat`)?

vanniktech:
> One option to make the dependency more explicit might be to hand in the emoji support lib config object to the GoogleCompatEmojiProvider constructor (and just make sure it's not null).

I think that's the best way to go for now.

The emoji support lib config object is available via the emoji library so we don't need to depend on the bundled one right? If so we should change this to be more explicit.

stefanhaustein:
Missing Emoji & Ctor parameter should be fixed now O:)

emoji-google-compat/src/main/java/com/vanniktech/emoji/googlecompat/GoogleCompatEmojiProvider.java
```diff
@@ -13,6 +14,13 @@
 import com.vanniktech.emoji.googlecompat.category.TravelCategory;
 
 public final class GoogleCompatEmojiProvider implements EmojiProvider {
+
+  public GoogleCompatEmojiProvider(@NonNull final EmojiCompat.Config emojiCompatConfig) {
+    if (emojiCompatConfig == null) {
+      throw new NullPointerException();
+    }
```
vanniktech: Let's also call the install function?

vanniktech:
Let's also call the install function?

stefanhaustein:
Distributing the initialization seems confusing to me. How about changing the ctor param from Config to EmojiCompat instead, i.e. the result of EmojiCompat.init(config);

vanniktech:
Let's do that I didn't know the init method returns something.

stefanhaustein:
Done! Merge? O:)

vanniktech:
Thanks a lot for all the good work :)

PR #197, opened 2017-08-30
Updated context for Nougat and newer devices

Debdeep009:
Changed the argument of `asActivity` inside the `EmojiPopup` constructor to pass the root view.
Changed the parameter of `asActivity()` in the `Utils` class to accept view instead of context.
Added code for Nougat and later for getting context.
For Nougat and later, `rootView.getContext()` returns an instance of `DecorContext` and not an `Activity` , so now the execution will not stop inside the while loop causing an anr.

### DIFF
--- a/emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
+++ b/emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
@@ -85,7 +85,7 @@ public final class EmojiPopup {
 
   EmojiPopup(@NonNull final View rootView, @NonNull final EmojiEditText emojiEditText,
              @Nullable final RecentEmoji recent, @Nullable final VariantEmoji variant) {
-    this.context = Utils.asActivity(rootView.getContext());
+    this.context = Utils.asActivity(rootView);
     this.rootView = rootView.getRootView();
     this.emojiEditText = emojiEditText;
     this.recentEmoji = recent != null ? recent : new RecentEmojiManager(context);
--- a/emoji/src/main/java/com/vanniktech/emoji/Utils.java
+++ b/emoji/src/main/java/com/vanniktech/emoji/Utils.java
@@ -6,6 +6,7 @@ import android.content.Context;
 import android.content.ContextWrapper;
 import android.graphics.Point;
 import android.graphics.Rect;
+import android.os.Build;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.view.View;
@@ -14,6 +15,7 @@ import android.widget.PopupWindow;
 
 import static android.os.Build.VERSION.SDK_INT;
 import static android.os.Build.VERSION_CODES.JELLY_BEAN;
+import static android.os.Build.VERSION_CODES.N;
 
 final class Utils {
   static final int DONT_UPDATE_FLAG = -1;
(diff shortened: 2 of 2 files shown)


vanniktech:
I'll look at this PR shortly. I'm rather busy at the moment.

Debdeep009:
I ran it on Samsung galaxy S7 and it was returning DecorContext. I think I still need to figure out why did that happen.

vanniktech:
@rubengees did this in #97 and AFAIR he took that method from the support library. Would be interesting to see whether they changed their implementation too or whether it's still the same.

vanniktech:
Any news?

Debdeep009:
I tested on a couple of Nougat devices and most of them are returning `DecorContext`. While still some of them aren't. Due to the multi-window feature, the `DecorView` has no longer access to the activity but the application as a whole. Take a look at the [setWindow()](https://android.googlesource.com/platform/frameworks/base.git/+/master/core/java/com/android/internal/policy/DecorView.java#1851) in `DecorView` class, so whenever `getContext()` returns DecorContext, the DecorContext has the phone window set. Also the extra api-check inside `asActivity()` isn't necessary.

Debdeep009:
Their [code](https://android.googlesource.com/platform/frameworks/support/+/2838eb1aea3fd1fac66526e1418f4739625bc4a1/v7/mediarouter/src/android/support/v7/app/MediaRouteButton.java#258) is still the same.

vanniktech:
So I guess we should leave it too?

Debdeep009:
Can you test it on any device - say google nexus 6p - v7.1.2 in genymotion and see as it returns `DecorContext`

Debdeep009:
I guess it is not safe to assume an context(esp. from rootView) would always return an Activity.

vanniktech:
I'll test it out tomorrow-ish, do I need to use the multi screen feature?

Debdeep009:
okay, no  prob! No just make sure to debug the context while testing on genymotion on Nougat devices.

vanniktech:
Will this also occur on normal devices or just when used together with Genymotion?

Debdeep009:
It occurs on Nougat and above - be it genymotion or hardware devices and I have seen that. My app was causing anr's on multiple devices due to the while check. Shall I send you a screenshot of the DecorContext while debugging ?

Debdeep009:
screenshot from 2017-09-13 01-24-19
Here have a look at the context.

vanniktech:
I ran Emoji on the normal AVD simulator image with API 25 and had no problems :/

Could it be that the layout is somehow different?

Debdeep009:
Which device image did you test on ? Was it a generic or ones from the genymotion ?

vanniktech:
Device: 2.7 QVGA Api 25

Which one are you using?

I'm still not that convinced that we need this change especially if the support lib still has the same code as we do plus.

Debdeep009:
Can you download and test on this image in genymotion - google nexus 6p - v7.1.2 for once. If it doesn't occur then this merge shouldn't be necessary.

vanniktech:
I don't have genymotion installed anymore. Also it's not the official simulator. The google ones have become pretty good these days and if on those devices you can find the same behavior feel free to reopen this one. Otherwise I'll close this. Also I've always had a few problems with the keyboard on Genymotion and decided to not use it for developing this, instead you can just use a normal android smartphone.

Debdeep009:
Okay will reopen this in case I face it on any other device:)

VitaliPom:
Hi, I'm new to github as a developer, so maybe I'll need some assistance to fix things.. 
I'm getting the exception thrown always on One Plus 5

VitaliPom:
Okay, just saw the debugger screenshot. I'll try to see what's hapenning inside for me and update

Debdeep009:
I guess that's what is caused due to the multi window mode

VitaliPom:
Gosh, so then I think I'll have to update it :) guys I'll ask your help.
Haven't committed a single line in here

VitaliPom:
Got it. Here is it: I am passing an actual activity and here if this is an activity, id doesn't even enter the while loop. Gotta try it, but maybe am I wrong? Anyone?

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

PR #198, opened 2017-08-30
Add infrastructure to let the provider perform emoji span replacements and utilize in emoji-google-compat

stefanhaustein:
Add an interface to enable the provider to do the spanning
Use support library spanning if available
Avoid empty array allocation 

Includes fixes from PR #199, which probably should go in first.

PR #199, opened 2017-08-31
Obey library loading state and add modifiers only where needed

stefanhaustein:
This is a strict subset of my other pull request, which also adds support library span handling. The purpose is to simplify the review process and to separate concerns.

It addresses two issues

- EmojiCompat span replacement routines are now only called when the library is ready. Before, an exception might be thrown if emojis were used at an earlier point of time.

- The emoji style modifler is added explicitly in the data where needed by the emoji support library. Before, it was added using an incomplete "rule of thumb" in the drawable. The previous approach was incorrect, overly broad and won't work if the whole text replacement is handled by the library.


vanniktech:
I'll look at this PR shortly. I'm rather busy at the moment.

vanniktech:
This has some of the things also from #198, right? If so I'd say let's focus on one PR first and let's do the other one. Feel free to choose with which we should start.

stefanhaustein:
This is a clean subset of #198, so ideally we'd do this first.

PR #208, opened 2017-09-20
Update Support Library version

### LINKED ISSUE #207: Outdated Support Library version
Version [`26.1.0`](https://developer.android.com/topic/libraries/support-library/revisions.html#26-1-0) of the Support Library had been released earlier this month. I think we should update this project to use that instead of the old `26.0.2`.

MrHadiSatrio:
This will fix #207.

Update the Support Library being used to version [`26.1.0`](https://developer.android.com/topic/libraries/support-library/revisions.html#26-1-0) released earlier this month.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -79,7 +79,7 @@ ext {
     targetSdkVersion = 26
     buildToolsVersion = '26.0.1'
 
-    supportLibraryVersion = '26.0.2'
+    supportLibraryVersion = '26.1.0'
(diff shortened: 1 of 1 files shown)


MrHadiSatrio:
I had been looking for ways to override before I made this PR but no luck – I kept on getting manifest merge error. Did I missed something?

vanniktech:
You can use gradle's resolution strategy.

Issue #88, opened 2017-02-14
Investigate how to bring the library size own


vanniktech:
PNGs are quite heavy for all of the emojis.

Possible solutions:
- vector drawables
- one big png image that contains all of the emojis

rubengees:
So I have found a source for Ios Emojis just now! WhatsApp has them in their Apk. Would it be fine to "steal" them?
We would have EmojiOne and Ios as Vectors then, I don't know about other variants yet.
The drawback would be a way less generic generator, but I think that is a low price for vectors 😁

shakil807g:
you can also use Telegram app approach they save single image of all the emojis and extract emojis from it

ArtworkAD:
@vanniktech so we could take the svg's from e.g. http://emojione.com/developers and convert them to vector drawables. Conversion is implemented with http://inloop.github.io/svg2android. This would reduce apk size and make the emojis scalable without loss of quality. We could introduce a new github repo that is automatically populated with vector drawable versions of svg emojis. What do your think?

Another approach would be to use a emoji font file. But I feel this would break your entire api

vanniktech:
I think that downloading the svgs and converting them to vector drawable could be implemented into the JavaScript generator module.

vanniktech:
Another option would be to generate one sprite PNG with all the emojis per category (apple, google, emojione) via the script and then extract the precise emoji within the library from the PNG.

stefanhaustein:
How about using the Emoji support library for the Google variant (or a separate Google variant)? It doesn't come with reasonable keyboard support and the current Google impl seems to duplicate the images in the case where one is also using the support library...?

stefanhaustein:
Can you take a look at my fork if this is heading in the right direction?
https://github.com/stefanhaustein/Emoji-1/tree/master/emoji-compat

vanniktech:
Looks pretty cool & interesting while glancing over it.

- I'd rename it to `emoji-google-compat`.
- It still needs to be wired together with the google support libraries or did I miss it?
- I saw that you ended up extending the Emoji class maybe a simple interface would be nice here so that you won't have to call those 4 super constructors.
- We could probably generate the Categories as well since we do that already for the normal google provider so adjusting those is not gonna be a lot of work.

Definitely thanks for starting to work on this.

stefanhaustein:
[...]

PR #252, opened 2018-02-04
Update emojis and use sprite sheet instead of individual images

rubengees:
This is a rewrite of the generator for updating the emojis and switching to the sprite sheet approach discussed in https://github.com/vanniktech/Emoji/issues/88.

We now use [this](https://github.com/iamcal/emoji-data) data source instead of the EmojiOne metadata file, which does not seem to get updated anytime soon.

I have marked this as WIP, as I'm not sure if this is the right way to go. I also have not regenerated yet to make this PR reviewable.

### Wins

- More emojis! :1st_place_medal:
- Size reduction. Not too great but notable. `Ios` went from 4.5MB to 3.7MB.

### Important code changes

A `<Name>Emoji` class is now generated for each emoji type (e.g. `IosEmoji`). This is a subclass of `Emoji` to make the sprite sheet approach work. It works by loading the big sheet into a static field and then cuts the individual emojis from that when requested.

The generator does not download any files anymore but depends on the aforementioned library through `npm` now.

### Points of discussion

- The ordering is different. It makes sense but is different from WhatsApp (which we always tried to copy in previous versions). It is now based on the official [unicode recommendations](https://www.unicode.org/emoji/charts/emoji-ordering.html). Slack uses this ordering now also for example.

- Once the big sheet is loaded, there is no way to reclaim that memory. Should we introduce a way of uninstalling a `Provider`? We then would also need a mechanism to destroy `Emoji` instances. Currently I'm thinking of making the `destroy` method of the `EmojiManager` public and adding a `destroy` method to `Emoji` which subclasses can override.


rubengees:
- The images are already heavily optimized ([pngquant](https://pngquant.org/) and [zopfli](https://github.com/google/zopfli)). That is the point of having one large image I think - You can optimize that pretty well.

- Yeah thought about that as well. We now have no `Provider` which uses the old logic from the `Emoji` class. I could imagine though that somebody is using that for a custom `Provider`. But I would be fine with converting to an interface.

- Sorry, totally forgot about that, here they are:
<details><summary>Click to expand</summary><p>

bildschirmfoto von 2018-02-04 13-29-08
bildschirmfoto von 2018-02-04 13-30-44 
bildschirmfoto von 2018-02-04 13-30-54
bildschirmfoto von 2018-02-04 13-31-04
bildschirmfoto von 2018-02-04 13-31-10
bildschirmfoto von 2018-02-04 13-31-17
bildschirmfoto von 2018-02-04 13-31-23
bildschirmfoto von 2018-02-04 13-31-30
bildschirmfoto von 2018-02-04 13-31-37
bildschirmfoto von 2018-02-04 13-31-47
bildschirmfoto von 2018-02-04 13-32-01
</p></details><br>

- Yes I actually have! In one app I work on, I constantly fight with OEMs and other problems related to low memory since that app is really image-heavvy. I would rather not have around 4MB loaded at all time but reclaim that memory when the user leaves the `Activity` with the emoji picker. I can imagine there are more people with similar problems...

vanniktech:
> The images are already heavily optimized (pngquant and zopfli). That is the point of having one large image I think - You can optimize that pretty well.

Nice.

> Yeah thought about that as well. We now have no Provider which uses the old logic from the Emoji class. I could imagine though that somebody is using that for a custom Provider. But I would be fine with converting to an interface.

Then leave things as is in this PR to minimize the changes.

> Sorry, totally forgot about that, here they are:

Oh wow things are really sorted differently. Is there any way to get a sorting that's close to Whatsapp ones? From the user perspective it's best if we use the same ordering otherwise you always need to end up looking for emojis which is the worst thing ever. Also we don't have any search functionality yet so this becomes even more important.

> Yes I actually have! In one app I work on, I constantly fight with OEMs and other problems related to low memory since that app is really image-heavvy. I would rather not have around 4MB loaded at all time but reclaim that memory when the user leaves the Activity with the emoji picker. I can imagine there are more people with similar problems...

Then let's do it.

rubengees:
> From the user perspective it's best if we use the same ordering otherwise you always need to end up looking for emojis which is the worst thing ever.

I agree! Sadly I don't know of a way to get the WhatsApp ordering with the data we have. We require a metadata file with categories and ordering (like that in WhatsApp) that currently only that file from EmojiOne has (at least I have not found another one). And that one does not get updated sadly.

Is this PR then out of question? I have looked around and there are at least two apps that have a somewhat similar ordering: Slack and the Telegram X app.

vanniktech:
> Is this PR then out of question?

No not at all. Let's ditch the order then and possibly try to tweak it. The benefits here outweigh that entirely.

rubengees:
I have added a machanism to reclaim memory. Here is a screenshot of it in action (Android Profiler):

bildschirmfoto von 2018-02-05 22-04-23

> possibly try to tweak it

I can't think of a non-manual approach, but will keep looking around.

emoji/src/main/java/com/vanniktech/emoji/EmojiManager.java
```diff
@@ -134,11 +134,16 @@ public static void install(@NonNull final EmojiProvider provider) {
     INSTANCE.emojiRepetitivePattern = Pattern.compile('(' + regex + ")+");
   }
 
-  static void destroy() {
+  public static void destroy() {
+    for (final Emoji emoji : INSTANCE.emojiMap.values()) {
```
vanniktech: Let's use fori here too

rubengees: We can't; [values()](https://docs.oracle.com/javase/7/docs/api/java/util/Map.html#values()) returns a `Collection` and not a `List`.

rubengees:
Adjusted a few things (the first two commits) and regenerated!

vanniktech:
Thanks. I'll take a look later.

PR #265, opened 2018-03-09
Add missing verifyInstalled to the EmojiManager

rubengees:
I just ran into a problem where this check could have avoided confusion, so I thought I quickly add it.

I also made `replaceWithImages` non-static like the other methods. It looks weird having the `EmojiManager` call `getInstance` on itself, that should be responsibility of the caller imo.


vanniktech:
Thanks

PR #266, opened 2018-03-09
Add release method which releases the sheet but not the data structure

rubengees:
This is from a personal use case which I can imagine is also relevant for others.

In my application, I want to release the heavy sheet `Bitmap` after leaving the relevant `Activity` but keep the emoji datastructure in memory to still have good performance when opening the `Activity` again. When using the existing `destroy` method, I would have to `install` a provider each time the `Activity` is opened, which blocks the main thread and causes a visible stutter in my case. This new `release` method allows to release the `Bitmap` when leaving and still installing the `Provider` once in a central place.


rubengees:
Sure! Give me a monent.

rubengees:
Done!

vanniktech:
Thanks once again :)

Issue #291, opened 2018-07-02
Project doesn't build as soon as we clone it


rocboronat:
Right now, the project asks to configure the `mavenPublish` task. May we offer a fake configuration to welcome collaborators that don't care about releasing versions?

`Error:Please configure repositoryUsername via mavenPublish.repositoryUsername or alternatively set the Gradle / System environment variable SONATYPE_NEXUS_USERNAME`

In addition, the project references `apply plugin: "com.vanniktech.maven.publish"` that breaks my head.

vanniktech:
Oh yes. Update that dependency to 0.4.0 please.

rocboronat:
Thanks for the quick answer :·)

vanniktech:
Sorry for the inconveniences.

rocboronat:
Wop, it also crashes.

You mean this dependence, isn't it?
`classpath 'com.vanniktech:gradle-maven-publish-plugin:0.4.0'`

PR #292, opened 2018-07-02
Add support to AutoCompleteTextView and MultiAutoCompleteTextView

rocboronat:
This PR just adds `EmojiAutoCompleteTextView` implementation. It's just a copypaste of the `EditText` class, but extending `AutoCompleteTextView` and using the new `attrs.xml` values.

BTW, as the project is not building right now, I didn't try it. When we fix it, I'll try it in the sample project.

Let's talk about it on #291


rocboronat:
@vanniktech feel free to ignore this PR till I notify you. Consider it a work in progress, just to showcase my progress... if we can consider it "progress", hehe!

vanniktech:
Is this ready now?

rocboronat:
@vanniktech if you like the approach I did to test both widgets, yes, it's ready to go. But I think we should remove the new Activities, because they exists only to test the new widgets, and didn't give too much value... What you think? If you like them, feel free to merge, and if not, I will remove last commit.

rocboronat:
Hey @vanniktech, do you have any thoughts about my answer?

rocboronat:
No problem, I'm happy to contribute to the project 😇

PR #293, opened 2018-07-02
Update Maven Publish Plugin to 0.4.0

### LINKED ISSUE #291: Project doesn't build as soon as we clone it

vanniktech:
Fixes #291


rocboronat:
Thanks! I'll check if it fixes the issue later :-)

Issue #191, opened 2017-08-17
Galaxy E7 Duos 3G SM-E700H : On this device the soft keyboard and emoji pop are inflated together.


DalveerSinghDaiya:
Thank you for filing an issue. If this is a bug that you want to report, please take the time to provide some information:

- Version of the library:
- Affected devices:
- Affected versions:

Screenshot(s) showing the issue and reprodution steps are appreciated.

----------------------------------------------------------------------

vanniktech:
Hmm do other applications that do this kind of thing work? E.g. WhatsApp or telegram?

DalveerSinghDaiya:
Nopes!! Its working fine in WhatsApp and Telegram.

vanniktech:
Hmmm :( that's not any good. I don't have the device at hand and in all the devices that I tested the library against (20+) devices. Are you willing to investigate this issue and maybe come up with a PR?

DalveerSinghDaiya:
I have tried some tricks but all for no good, its not working, that's why i filed it here, thought some one would have cracked this already.

vanniktech:
Yeah as you probably know detecting whether or not the keyboard shown is non trivial especially when certain devices / manufacters do their own thing. Maybe someone will crack it.

DalveerSinghDaiya:
Haha, I was expecting you to do it :P

vanniktech:
Nope I'm also not actively looking into it. Feel free to take this one

vanniktech:
Related to #224.

@DalveerSinghDaiya could you try the possible workarounds that were posted in that issue and see if that would resolve your issue?

daniele-athome:
Despite I would really prefer to keep the current behavior, I would like to point out that Hangouts (meaning it's Google doing this) closes the keyboard before showing the emoji panel. It seems like they gave up eheh :-)
If you need help to test or code something to fix this, I'm here.

vanniktech:
We need to show the keyboard in order to get the height.

daniele-athome:
> We need to show the keyboard in order to get the height.

Sure, I mean get the height and then close it. I believe that's what Hangouts does.

vanniktech:
That sounds good. Happy to review an implementation for that.

daniele-athome:
[...]

PR #306, opened 2018-10-02
fix keyboard with customize keyboard

chihung93:
Customize for fixing emoji on some specialist devices

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -14,7 +14,9 @@
         tools:ignore="GoogleAppIndexingWarning">
         <activity
             android:name=".MainActivity"
-            android:windowSoftInputMode="stateHidden">
+            android:windowSoftInputMode="adjustResize"
+
+            >
             <intent-filter>
                 <action android:name="android.intent.action.MAIN"/>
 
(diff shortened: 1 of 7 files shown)


vanniktech:
Thanks for starting to work on this. Can anyone who's gotten the issue from #191 verify that this does fix the issue as I can't reproduce? @daniele-athome @danielgaribaldi @186538

daniele-athome:
I could try with [my user's way of reproducing the issue](https://github.com/kontalk/androidclient/issues/1110#issuecomment-348548271). I'll let you know. Thanks @chihung93!

vanniktech:
@daniele-athome any news?

daniele-athome:
Sorry for being late. I've been testing the sample app, the close-keyboard-before-showing-emoji mechanism works, but there is something wrong with the detection of the keyboard height, so the EditText moves up and down between keyboard and emojis. I might do something to fix that (I already do it in my code), but I have a couple of questions:

* EditText cursor disappears because the view loses focus (it shouldn't lose it. Right?)
* I guess the keyboard height is not handled automatically by your view but it has to be provided via setKeyboardHeight
* AFAICU the emoji panel is hidden when the EditText is clicked. I feel like it's wrong. OnClick might not be the only event that triggers the keyboard (what about OnFocus? Or anything else)

I feel I can somehow contribute to this, but I can't promise anything on short term. Anyway I'll have to work on this sooner or later :-)

vanniktech:
Regarding your first question yes.
Yeah custom keyboard height needs to be considered. We might even need to special case the notch on some devices.
OnFocus like selecting the text should also dismiss it. 

Thanks for picking this up in the future!

vanniktech:
This is super old now and in a state on which we can't advance further. Hence I'm closing this but willing to reconsider once the PR reaches a mergeable state.

PR #312, opened 2018-10-25
use EditText for builder popup instead of EmojiEditText

sebest06:
Hi, this is the change that I suggest
Thank

### DIFF
--- a/emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
+++ b/emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
@@ -11,9 +11,11 @@ import android.support.annotation.ColorInt;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.view.Gravity;
+import android.view.KeyEvent;
 import android.view.View;
 import android.view.ViewTreeObserver;
 import android.view.inputmethod.InputMethodManager;
+import android.widget.EditText;
 import android.widget.PopupWindow;
 import com.vanniktech.emoji.emoji.Emoji;
 import com.vanniktech.emoji.listeners.OnEmojiBackspaceClickListener;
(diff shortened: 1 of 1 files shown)


vanniktech:
A change for what? What's your motivation and what is this trying to solve?

sebest06:
Hi how are you?

For example you have an application with support for "google-compat" to show the emoticons correctly, but the only way to enter them is through the android keyboard. The emoticons look different on the keyboard than what will be displayed.

Here I can integrate your library with the application, if the object (Edit) that I use in your library (to the build method) could be a TextEdit I can integrate your library without problems, my object is not a TextEdit, because it is compatible with google -compat, but inherit from a TextEdit.

In the current version I should implement the EmojiEditText interface, in my Edit type object, and overload some functions, to do something I'm already doing.

Even with the modification that I propose, you could still pass an EmojiEditText to the build method and the library would continue behaving as it does now.

If you look at the code I sent you, I just modify the type, but the rest of the example code is the same.

In this way you could have unlinked the keyboard of the object that receives the emoticon. Which makes it simpler to integrate in different scenarios.

I think it's a change that adds flexibility and does not take anything away. No matter what type of object you use to enter information you can always receive data from your emoticon keyboard.

vanniktech:
Ah okay. I think I got it. This will require probably some changes to our EditText too. I'll have a deeper look hopefully next week.

I think what needs to be done is inlining the backspace / input method. As well as removing the interface.

premacck:
Hey any updates on this PR?

vanniktech:
Haven't had time to have a deeper look at this.

sebest06:
Ok i know that. I hope we can do that the next year... Happy xmas.

sebest06:
Hello Vanniktech, how are you? i know that you are busy, but i want to say somethings

I integreted your library in two big projects.
One of these is not happy with the implementation i do not know why...

The other one was more happy, and he asked me a few things of the implementation.

https://github.com/kriztan/Pix-Art-Messenger/pull/288

This is the pull request for if you want to read.

I said that the library is yours and what would this tell you.

Could you take a look when you have time?

Thanks!

vanniktech:
Still on my list. Maybe this week is the week.

PR #320, opened 2018-12-17
Add attrs for color customization on theme level

rubengees:
This allows to set these colors also on a theme level.
This change is backwards compatible, since the old behaviour is used, if an attribute is not defined in the theme.


vanniktech:
CI would need to be fixed and then feel free to merge (added you as a collaborator)

rubengees:
Cool, thanks! Since CI is green, I'll go ahead and merge this :slightly_smiling_face:

PR #326, opened 2019-01-21
Fix Layout issues in EmojiDialog for several edge cases.

mario:
This PR fixes a few layout issues I've noticed, more importantly:

- it now properly aligns the emoji keyboard above soft navigation buttons at the bottom on some phones
- it now properly shows the emoji keyboard when using default edit text behaviour in horizontal orientation

This is still WIP as I want to clean it up and test, but feel free to test it yourself and comment - all feedback is more than welcome.

### DIFF
--- a/emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
+++ b/emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
@@ -16,6 +16,7 @@ import android.view.ViewTreeObserver;
 import android.view.inputmethod.InputMethodManager;
 import android.widget.EditText;
 import android.widget.PopupWindow;
+
 import com.vanniktech.emoji.emoji.Emoji;
 import com.vanniktech.emoji.listeners.OnEmojiBackspaceClickListener;
 import com.vanniktech.emoji.listeners.OnEmojiClickListener;
@@ -30,6 +31,7 @@ import static com.vanniktech.emoji.Utils.checkNotNull;
 
 public final class EmojiPopup {
   static final int MIN_KEYBOARD_HEIGHT = 100;
+  static final float HEIGHT_DIFFERENCE_FACTOR = 1.4f;
 
   final View rootView;
   final Activity context;
@@ -52,23 +54,22 @@ public final class EmojiPopup {
   @Nullable OnEmojiClickListener onEmojiClickListener;
   @Nullable OnEmojiPopupDismissListener onEmojiPopupDismissListener;
 
-  int correctionFactor = 0;
+  int correctionFactor;
 
   final ViewTreeObserver.OnGlobalLayoutListener onGlobalLayoutListener = new ViewTreeObserver.OnGlobalLayoutListener() {
     @Override public void onGlobalLayout() {
       final Rect rect = Utils.windowVisibleDisplayFrame(context);
       final int heightDifference = Utils.getScreenHeight(context) - rect.bottom;
 
-      boolean shouldOverrideRegularCondition =
-          Utils.shouldOverrideRegularCondition(context, editText);
-      if ((heightDifference > Utils.dpToPx(context, MIN_KEYBOARD_HEIGHT)
-          || shouldOverrideRegularCondition)) {
+      final boolean shouldOverrideRegularCondition = Utils.shouldOverrideRegularCondition(context, editText);
 
+      if (heightDifference > Utils.dpToPx(context, MIN_KEYBOARD_HEIGHT) || shouldOverrideRegularCondition) {
         correctionFactor = rect.top;
 
         int height = 0;
+
         if (shouldOverrideRegularCondition) {
-          height = (int) ((Utils.getScreenHeight(context) / 2) - (heightDifference * 1.4));
+          height = (int) ((Utils.getScreenHeight(context) / 2) - (heightDifference * HEIGHT_DIFFERENCE_FACTOR));
           popupWindow.setHeight(height);
         } else {
           height = heightDifference + correctionFactor;
@@ -99,8 +100,7 @@ public final class EmojiPopup {
           }
 
           dismiss();
-          Utils.removeOnGlobalLayoutListener(context.getWindow().getDecorView(),
-              onGlobalLayoutListener);
+          Utils.removeOnGlobalLayoutListener(context.getWindow().getDecorView(), onGlobalLayoutListener);
         }
       }
     }
--- a/emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
+++ b/emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
@@ -29,7 +29,7 @@ import static com.vanniktech.emoji.Utils.backspace;
 import static com.vanniktech.emoji.Utils.checkNotNull;
 
 public final class EmojiPopup {
-  private static final int MIN_KEYBOARD_HEIGHT = 100;
+  static final int MIN_KEYBOARD_HEIGHT = 100;
 
   final View rootView;
   final Activity context;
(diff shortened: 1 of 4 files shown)


mario:
@vanniktech well, I can try :) Would appreciate testing in the mean time!

vanniktech:
It behaves very weirdly on my emulator.

screenshot_1548063558

Also toggling with the Emoji button no longer seems to work.

mario:
Can you let me know which emulator is that so I can take a look? (what Android version)

vanniktech:
`9d477d27` still behaves flaky with the same behavior.

mario:
Yea, I just did reformatting to comply with your rules. The code did work, so maybe I didn't push the right branch - I'm investigating.

mario:
Should be ready now. I've experienced some weird "dismiss" animation issues but those seem to be there before as well.

This fixed extracted UI issues + issues with phones like P20 pro.

mario:
@rubengees I now use the exact same style as you folks do (Square Android) and the formatting has been done according to those guidelines. 

Can you clarify what should I change? Do you mean the private modifiers I've added and such?

rubengees:
No, those are fine! Things like these:

emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
```diff
@@ -99,13 +128,16 @@
...
+      public void onEmojiLongClick(@NonNull final EmojiImageView view, @NonNull final Emoji emoji) {
         variantPopup.show(view, emoji);
       }
     };
 
     final OnEmojiClickListener clickListener = new OnEmojiClickListener() {
-      @Override public void onEmojiClick(@NonNull final EmojiImageView imageView, @NonNull final Emoji emoji) {
+      @Override
+      public void onEmojiClick(@NonNull final EmojiImageView imageView,
+          @NonNull final Emoji emoji) {
```
rubengees: Here for example.

emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
```diff
@@ -122,9 +154,12 @@
 
     variantPopup = new EmojiVariantPopup(this.rootView, clickListener);
 
-    final EmojiView emojiView = new EmojiView(context, clickListener, longClickListener, recentEmoji, variantEmoji, backgroundColor, iconColor, dividerColor);
+    final EmojiView emojiView =
+        new EmojiView(context, clickListener, longClickListener, recentEmoji, variantEmoji,
+            backgroundColor, iconColor, dividerColor);
```
rubengees: Or this. When first looking at this, it suggests an actual change has been made to the `EmojiView` constructor or it's parameters.

emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
mario: Ok, I imagine this is because of chars limit in a line. Will try to get to it soon!

emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
mario: All good, I understand the pain.

emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
rubengees: Great, thanks!

mario: Can you let me know the char limit per line that you folks use?

rubengees: Personally I use 120, but this project doesn't use one I think, the lines can be as long as they want (At least I have seen lines longer then my configured limit in the IDE). I think it would be good to introduce a char limit (and update the style in general), but that should be a separate PR.

mario:
So hard on the new contributors :P

Anyway, I tried ... let me know if I missed something, it's getting quite late :)

emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
```diff
@@ -30,66 +29,85 @@
 import static com.vanniktech.emoji.Utils.checkNotNull;
 
 public final class EmojiPopup {
-  static final int MIN_KEYBOARD_HEIGHT = 100;
+  private static final int MIN_KEYBOARD_HEIGHT = 100;
```
vanniktech: Please leave all of these package-private. They are so that we don't create synthetic accessor methods.

emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
```diff
@@ -30,66 +29,85 @@
...
+
+  @Nullable private OnEmojiBackspaceClickListener onEmojiBackspaceClickListener;
+  @Nullable private OnEmojiClickListener onEmojiClickListener;
+  @Nullable private OnEmojiPopupDismissListener onEmojiPopupDismissListener;
+
+  private int correctionFactor = 0;
+
+  private final ViewTreeObserver.OnGlobalLayoutListener onGlobalLayoutListener =
+      new ViewTreeObserver.OnGlobalLayoutListener() {
+        @Override
```
vanniktech: nit: this should on the same line as the method

emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
```diff
@@ -30,66 +29,85 @@
...
         }
-      }
-    }
-  };
+      };
 
   EmojiPopup(@NonNull final View rootView, @NonNull final EditText editText,
-            @Nullable final RecentEmoji recent, @Nullable final VariantEmoji variant,
-            @ColorInt final int backgroundColor, @ColorInt final int iconColor, @ColorInt final int dividerColor) {
+      @Nullable final RecentEmoji recent, @Nullable final VariantEmoji variant,
```
vanniktech: Nice!

emoji/src/main/java/com/vanniktech/emoji/Utils.java
```diff
@@ -48,12 +50,19 @@ static int dpToPx(@NonNull final Context context, final float dp) {
...
+      return (configuration.orientation == Configuration.ORIENTATION_LANDSCAPE);
+    }
 
-    return size.y;
+    return false;
+  }
+
+  static int screenHeight(@NonNull final Activity context) {
+    Configuration configuration = context.getResources().getConfiguration();
+    return (dpToPx(context, configuration.screenHeightDp));
```
vanniktech: nit: the first `(` & `)` are unnecessary

mario: They might be, but I like them. But given it's your guidelines, I'll remove them.

mario:
@vanniktech can you let me know where (what phone, what OS version) do you experience the glitch with the bottom part not filled?

As for #191, I believe so - I'll take a look at my S7 - might not be the EXACT same phone, but oh well.

vanniktech:
(I can also take care of the formatting in case it's too exhausting for you.) We'll get there slowly.

Emulator is Nexus 4 with API 28. And on my Pixel 3 this also fails with the same behavior which is on Android P too.

mario:
@vanniktech I plan on contributing in the long term unless something serious happens, so I'd rather slowly familiarize myself with the guidelines and all. Luckily, I've been in FOSS for my whole life so I don't find the approach off-putting (and understand the reasoning), but I can understand how a new contributor might want to get away after so many style requests on their first PR :P

vanniktech:
Weird. Not that I know. I do think that the notch isn't supported which is why it's failing on my Pixel 3XL.

mario:
@vanniktech my P20 Pro has a notch, and it works as expected. Can you try to clean & rebuild? Probably not it, but you never know.

mario:
Ah missed the two, sorry about that. Will try on a few more devices, but it's really strange that your Nexus 4 emulator (which is the same as mine!) behaves differently.

mario:
I just simulated a cutout too (both regular & double) and it works. Uh.

mario:
Ah found a way to reproduce. Will fix.

mario:
Can you folks try it now please?

vanniktech:
Thank you very much!

mario:
Thanks fo both of you for testing and doing code review.

rubengees:
Thanks for this change @mario! I was not in time to review, but I integrated this commit in one of my apps and confirmed with a user having a similar problem, that it now works!

vanniktech:
If you want @rubengees next time I'll wait for your review too.

PR #329, opened 2019-01-24
Fix Emoji keyboard in certain cases

### LINKED ISSUE #328: Weird behaviour in landscape when flagNoExtractUi is not set
- Version of the library: 748c208
- Affected devices: All
- Affected versions: All

When opening the emoji keyboard in landscape, the emoji picker is shown on it's own, without the `EditText`. I reproduced that on a default API 28 emulator.
Setting `android:imeOptions="flagNoExtractUi"` works around like before. 

@mario I think that is related to your change?

mario:
This fixes an issue brought to my attention in #328 and a few other issues I've noticed with the Popup view.

### DIFF
--- a/app/src/main/java/com/vanniktech/emoji/sample/MainActivity.java
+++ b/app/src/main/java/com/vanniktech/emoji/sample/MainActivity.java
@@ -114,11 +114,7 @@ import com.vanniktech.emoji.twitter.TwitterEmojiProvider;
   }
 
   @Override public void onBackPressed() {
-    if (emojiPopup != null && emojiPopup.isShowing()) {
-      emojiPopup.dismiss();
-    } else {
-      super.onBackPressed();
-    }
+    super.onBackPressed();
   }
 
   @Override protected void onStop() {
--- a/app/src/main/java/com/vanniktech/emoji/sample/MainActivity.java
+++ b/app/src/main/java/com/vanniktech/emoji/sample/MainActivity.java
@@ -113,10 +113,6 @@ import com.vanniktech.emoji.twitter.TwitterEmojiProvider;
     }
   }
 
-  @Override public void onBackPressed() {
-    super.onBackPressed();
-  }
-
   @Override protected void onStop() {
     if (emojiPopup != null) {
       emojiPopup.dismiss();
--- a/emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
+++ b/emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
@@ -197,15 +197,19 @@ public final class EmojiPopup {
 
   private void showAtBottomPending() {
     isPendingOpen = true;
-    final InputMethodManager inputMethodManager =
-        (InputMethodManager) context.getSystemService(Context.INPUT_METHOD_SERVICE);
+    InputMethodManager inputMethodManager = (InputMethodManager) context.getSystemService(Context.INPUT_METHOD_SERVICE);
 
     if (Utils.shouldOverrideRegularCondition(context, editText)) {
       editText.setImeOptions(editText.getImeOptions() | EditorInfo.IME_FLAG_NO_EXTRACT_UI);
-      inputMethodManager.restartInput(editText);
+      if (inputMethodManager != null) {
+        inputMethodManager.restartInput(editText);
+      }
     }
 
-    inputMethodManager.showSoftInput(editText, InputMethodManager.RESULT_UNCHANGED_SHOWN, resultReceiver);
+    if (inputMethodManager != null) {
+      inputMethodManager.showSoftInput(editText, InputMethodManager.RESULT_UNCHANGED_SHOWN, resultReceiver);
+      inputMethodManager = null;
+    }
   }
 
   public boolean isShowing() {
@@ -219,10 +223,12 @@ public final class EmojiPopup {
     variantEmoji.persist();
 
     if (originalImeOptions != -1) {
-      final InputMethodManager inputMethodManager =
-          (InputMethodManager) context.getSystemService(Context.INPUT_METHOD_SERVICE);
+      InputMethodManager inputMethodManager = (InputMethodManager) context.getSystemService(Context.INPUT_METHOD_SERVICE);
       editText.setImeOptions(originalImeOptions);
-      inputMethodManager.restartInput(editText);
+      if (inputMethodManager != null) {
+        inputMethodManager.restartInput(editText);
+        inputMethodManager = null;
+      }
     }
   }
 
--- a/emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
+++ b/emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
@@ -1,17 +1,19 @@
 package com.vanniktech.emoji;
 
+import android.annotation.TargetApi;
 import android.app.Activity;
 import android.content.Context;
 import android.graphics.Bitmap;
-import android.graphics.Point;
 import android.graphics.Rect;
 import android.graphics.drawable.BitmapDrawable;
+import android.os.Build;
 import android.support.annotation.CheckResult;
 import android.support.annotation.ColorInt;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.support.annotation.StyleRes;
 import android.support.v4.view.ViewPager;
+import android.util.Log;
 import android.view.Gravity;
 import android.view.View;
 import android.view.ViewTreeObserver;
@@ -26,12 +28,15 @@ import com.vanniktech.emoji.listeners.OnEmojiPopupDismissListener;
 import com.vanniktech.emoji.listeners.OnEmojiPopupShownListener;
 import com.vanniktech.emoji.listeners.OnSoftKeyboardCloseListener;
 import com.vanniktech.emoji.listeners.OnSoftKeyboardOpenListener;
+import java.lang.reflect.Field;
 
 import static com.vanniktech.emoji.Utils.backspace;
 import static com.vanniktech.emoji.Utils.checkNotNull;
 import static com.vanniktech.emoji.Utils.shouldOverrideRegularCondition;
 
 public final class EmojiPopup {
+  static final String TAG = "EmojiPopup";
+
   static final int MIN_KEYBOARD_HEIGHT = 100;
 
   final View rootView;
@@ -55,57 +60,74 @@ public final class EmojiPopup {
   @Nullable OnEmojiClickListener onEmojiClickListener;
   @Nullable OnEmojiPopupDismissListener onEmojiPopupDismissListener;
 
-  int correctionFactor;
+  int viewInset;
+  Rect rect = new Rect();
 
   final ViewTreeObserver.OnGlobalLayoutListener onGlobalLayoutListener = new ViewTreeObserver.OnGlobalLayoutListener() {
     @Override @SuppressWarnings("PMD.CyclomaticComplexity") public void onGlobalLayout() {
-      final Rect rect = Utils.windowVisibleDisplayFrame(context);
-      final int heightDifference = Utils.getScreenHeight(context) - rect.bottom;
-
-      final boolean shouldOverrideRegularCondition = Utils.shouldOverrideRegularCondition(context, editText);
-
-      if (heightDifference > Utils.dpToPx(context, MIN_KEYBOARD_HEIGHT) || shouldOverrideRegularCondition) {
-        correctionFactor = rect.top;
-
-        int height = 0;
-
-        if (shouldOverrideRegularCondition) {
-          height = Utils.getScreenHeight(context) - editText.getHeight() - correctionFactor ;
-          popupWindow.setHeight(height);
-        } else {
-          height = heightDifference + correctionFactor;
-          popupWindow.setHeight(height);
-        }
-        popupWindow.setWidth(rect.right);
-
-        if (!isKeyboardOpen && onSoftKeyboardOpenListener != null) {
-          onSoftKeyboardOpenListener.onKeyboardOpen(height);
-        }
-
-        isKeyboardOpen = true;
-
-        if (isPendingOpen) {
-          showAtBottom();
-          isPendingOpen = false;
-        }
-      } else {
-        if (heightDifference < 0) {
-          correctionFactor = heightDifference;
-        }
-
-        if (isKeyboardOpen) {
-          isKeyboardOpen = false;
-
-          if (onSoftKeyboardCloseListener != null) {
-            onSoftKeyboardCloseListener.onKeyboardClose();
-          }
-
-          dismiss();
-          Utils.removeOnGlobalLayoutListener(context.getWindow().getDecorView(), onGlobalLayoutListener);
-        }
-      }
+      updateKeyboardState();
     }
   };
+
+  @TargetApi(Build.VERSION_CODES.LOLLIPOP)
+  private int getViewInset() {
+    try {
+      Field attachInfoField = View.class.getDeclaredField("mAttachInfo");
+      attachInfoField.setAccessible(true);
+      Object attachInfo = attachInfoField.get(rootView);
+      if (attachInfo != null) {
+        Field stableInsetsField = attachInfo.getClass().getDeclaredField("mStableInsets");
+        stableInsetsField.setAccessible(true);
+        Rect insets = (Rect) stableInsetsField.get(attachInfo);
+        return insets.bottom;
+      }
+    } catch (NoSuchFieldException nsfe) {
+      Log.w(TAG, "Field error when measuring view inset", nsfe);
+    } catch (IllegalAccessException iae) {
+      Log.w(TAG, "Illegal Access reflection error when measuring view inset", iae);
+    }
+
+    return 0;
+  }
+
+  private void updateKeyboardState() {
+    if (viewInset == 0 && Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+      viewInset = getViewInset();
+    }
+
+    rect = Utils.windowVisibleDisplayFrame(context);
+
+    final int availableHeight = rootView.getHeight() - viewInset - (!Utils.isFullScreen(context) ? rect.top : 0);
+    final int keyboardHeight = Utils.shouldOverrideRegularCondition(context, editText) ?
+        Utils.getScreenHeight(context) - editText.getBottom() - (!Utils.isFullScreen(context) ? rect.top : 0)
+        : availableHeight - (rect.bottom - rect.top);
+
+    if (keyboardHeight > Utils.dpToPx(context, MIN_KEYBOARD_HEIGHT)) {
+      if (popupWindow.getHeight() != keyboardHeight) {
+        popupWindow.setHeight(keyboardHeight);
+      }
+
+      if (popupWindow.getWidth() != rect.right) {
+        popupWindow.setWidth(rect.right);
+      }
+
+      if (!isKeyboardOpen) {
+        isKeyboardOpen = true;
+        if (onSoftKeyboardOpenListener != null) {
+          onSoftKeyboardOpenListener.onKeyboardOpen(keyboardHeight);
+        }
+      }
+
+      if (isPendingOpen) {
+        showAtBottom();
+      }
+    } else if (isKeyboardOpen) {
+      isKeyboardOpen = false;
+      if (onSoftKeyboardCloseListener != null) {
+        onSoftKeyboardCloseListener.onKeyboardClose();
+      }
+    }
+  }
 
   EmojiPopup(@NonNull final View rootView, @NonNull final EditText editText,
       @Nullable final RecentEmoji recent, @Nullable final VariantEmoji variant,
@@ -156,7 +178,8 @@ public final class EmojiPopup {
 
     popupWindow.setContentView(emojiView);
     popupWindow.setInputMethodMode(PopupWindow.INPUT_METHOD_NOT_NEEDED);
-    popupWindow.setBackgroundDrawable(new BitmapDrawable(context.getResources(), (Bitmap) null)); // To avoid borders and overdraw.
+    popupWindow.setBackgroundDrawable(
+        new BitmapDrawable(context.getResources(), (Bitmap) null)); // To avoid borders and overdraw.
     popupWindow.setOnDismissListener(new PopupWindow.OnDismissListener() {
       @Override public void onDismiss() {
         if (onEmojiPopupDismissListener != null) {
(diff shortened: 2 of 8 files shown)


emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
```diff
@@ -172,18 +174,22 @@ public void toggle() {
 
       if (isKeyboardOpen) {
         // If the keyboard is visible, simply show the emoji popup.
-        showAtBottom();
+        editText.postDelayed(new Runnable() {
+          @Override public void run() {
+            showAtBottom();
+          }
+        }, 50);
```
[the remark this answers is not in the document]
mario: Because in rare cases, due to race conditions, we get a smaller height from the keyboard than expected (because it's not fully shown), so the emoji popvuiew is smaller than needed.

emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
```diff
@@ -212,13 +219,21 @@ void showAtBottom() {
     popupWindow.showAtLocation(rootView, Gravity.NO_GRAVITY, desiredLocation.x, desiredLocation.y);
 
     if (onEmojiPopupShownListener != null) {
+      if (shouldOverrideRegularCondition(context, editText)) {
+        final InputMethodManager inputMethodManager = (InputMethodManager) context.getSystemService(Context.INPUT_METHOD_SERVICE);
+        inputMethodManager.showSoftInput(editText, 0);
```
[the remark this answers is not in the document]
mario: Because the keyboard won't open in landscape mode if you don't have NoExtractUI flag set, so we have to do it this way.

mario:
I'll investigate the sizing when NOT using adjustResize.

emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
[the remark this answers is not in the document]
rubengees: Did it before the other PR of yours?

mario:
Can you let me know the combination you use for activity + edittext (imeOptions) so I can try to debug as I can't really reproduce?

mario:
@rubengees when you get the time to answer, I'll take a look. Got another PR in the mean time :)

rubengees:
On my `Activity` I have no special properties (so no `adjustResize`), the `EmojiEditText` is like this (simplified):

```xml
<com.vanniktech.emoji.EmojiEditText
    android:layout_width="match_parent"
    android:inputType="textMultiLine|textCapSentences|textAutoCorrect"
    android:maxLines="5"
    android:scrollbars="vertical" />
```

mario:
I now came up with a significantly better solution and have rebased the PR against master. Please test and reply accordingly - thanks!

vanniktech:
I'll test it out on Monday. Are there any further changes you want to see for 0.6.0?

vanniktech:
Nah. Just GitHub. Maybe Ruben can help with the Emojis. Also what should we do about EmojiOne?

mario:
I'd drop it if that's what the folks prefer.

vanniktech:
Can we then also update the Emoji ordering @rubengees? I feel like when cutting a new release we should have all of the emojis in place.

rubengees:
Regarding this PR - still does not work in landscape for me. The popup is not shown at all...

> Maybe Ruben can help with the [ignored] Emojis.

I'll try to look into that 🙂 

> Can we then also update the Emoji ordering?

You mean the order in which the emojis are shown in the picker? If yes, I can't think of a way. The generator simply uses the ordering of our datasource.

mario:
@rubengees can you talk the Sample app shipped in the repo (and this PR, in particular) to exhibit the behaviour you mentioned so I can properly debug? I've tried a few combinations and they all work as expected for me :(

rubengees:
@mario Hmm it seems to have been an issue with my app, at least partially. I hide and show the input based on data that is loaded and also instantiated the popup only when it was first used (using Kotlin's `lazy`) and that seemed to have put the logic into an awkward state.

It works mostly now. But I have noticed this:




The first screenshot shows the landscape view without the picker opened. The second with it opened. As you can see, the "done" button is not accessible for the user anymore. I think the popup should only have the height of the keyboard in landscape also.

mario:
So the suggestion is to instantiate the popup as soon as possible due to global layout listener in general. Let me investigate your issue now :)

mario:
Nop, can't reproduce - can you send me a sample layout file? I just remove all the adjustResize and set the flagNoExtractUI and it works as expected :(

rubengees:
I am currently on mobile only, but will try to put something together and investigate more thoroughly asap. My observations above were only quick findings.

mario:
Thanks!

mario:
Sorry to bother, but any progress on getting me a way to reproduce or a fix @rubengees?

rubengees:
Sorry, not yet. I'm currently busy with work, but have not forgotten about it!

rubengees:
I have investigated further and the problems with the popup not showing were all linked to problems with my app, sorry for the confusion... Maybe we should add a disclaimer to the README, that the popup needs to be instantiated as early as possible, to avoid others to make the same mistake? 

Apart from that, only the problem with the popup being above the "done" button is left for me. I can reproduce that without any modifications to the sample or the library. I happens on my real phone (OnePlus 5T, Android 9.0 stock) and also on an default Android Studio emulator (Pixel 2, Android 9.0)

mario:
> I have investigated further and the problems with the popup not showing were all linked to problems with my app, sorry for the confusion... Maybe we should add a disclaimer to the README, that the popup needs to be instantiated as early as possible, to avoid others to make the same mistake?
> 
> Apart from that, only the problem with the popup being above the "done" button is left for me. I can reproduce that without any modifications to the sample or the library. It happens on my real phone (OnePlus 5T, Android 9.0 stock) and also on an default Android Studio emulator (Pixel 2, Android 9.0).

Well, definitely good news! As for the second thing - would you mind telling me where in the sample (either in master or this PR) do you see the "Done" button? :-/ I must be missing something.

rubengees:
@mario Here:




In the first screenshot, you can see the "send" button (in my app, it's a "done" button, but that doesn't make a difference). In the second screenshot, it cannot be accessed, due to the popup being over it.

mario:
Thanks. Will try that emulator then.

mario:
Took me forever to figure this one out, but I guess this is why there are no solutions to this available so far :(

Please test it.

emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
```diff
@@ -97,9 +97,14 @@ private void updateKeyboardState() {
...
 
     final int availableHeight = rootView.getHeight() - viewInset - (!Utils.isFullScreen(context) ? rect.top : 0);
-    final int keyboardHeight = Utils.shouldOverrideRegularCondition(context, editText) ?
-        Utils.getScreenHeight(context) - editText.getBottom() - (!Utils.isFullScreen(context) ? rect.top : 0)
-        : availableHeight - (rect.bottom - rect.top);
+
+    int keyboardHeight;
+
+    if (Utils.shouldOverrideRegularCondition(context, editText)) {
+      keyboardHeight = (int) ((availableHeight - editText.getBottom() - (0.252 * availableHeight)) + (!Utils.isFullScreen(context) ? availableHeight * 0.055 : 0));
```
rubengees: Can you explain what is going on here?

mario:
Absolutely, if you can test and confirm it works while I rebase? :)

mario:
By how much? Can you screenshot for me?

mario:
Simplified the code and fixed #340 along the way. One more bug remains with landscape orientation that I'm trying to hunt down.

emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
mario: It is a magic calculation based off of some fractions I've found in the Android source code. Not SUPER reliable, but it seemed to work on all the phones I've had. Can you show me a screenshot on how it looks like on yours?

mario:
I *finally* got it. I've opted to show emoji keyboard in a regular view instead of a full screen view because that's how a few other IM apps do it, but it's absolutely the same size as the regular keyboard you use.

Please test, merge, and let's do a release!

app/src/main/java/com/vanniktech/emoji/sample/MainActivity.java
```diff
@@ -114,11 +114,7 @@
   }
 
   @Override public void onBackPressed() {
-    if (emojiPopup != null && emojiPopup.isShowing()) {
-      emojiPopup.dismiss();
-    } else {
-      super.onBackPressed();
-    }
+    super.onBackPressed();
```
vanniktech: If we don't do anything here any longer then we can remove the method completely.

emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
```diff
@@ -55,57 +59,48 @@
   @Nullable OnEmojiClickListener onEmojiClickListener;
   @Nullable OnEmojiPopupDismissListener onEmojiPopupDismissListener;
 
-  int correctionFactor;
+  int originalIMEOptions = -1;
```
```suggestion
  int originalImeOptions = -1;
```

emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
```diff
@@ -55,57 +59,48 @@
...
+      updateKeyboardState();
+    }
+  };
 
-      final boolean shouldOverrideRegularCondition = Utils.shouldOverrideRegularCondition(context, editText);
+  private void updateKeyboardState() {
 
-      if (heightDifference > Utils.dpToPx(context, MIN_KEYBOARD_HEIGHT) || shouldOverrideRegularCondition) {
-        correctionFactor = rect.top;
+    int keyboardHeight = Utils.getInputMethodHeight(context);
```
```suggestion
    final int keyboardHeight = Utils.getInputMethodHeight(context);
```

emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
```diff
@@ -55,57 +59,48 @@
...
 
-        if (shouldOverrideRegularCondition) {
-          height = (int) (Utils.getScreenHeight(context) / 2f - heightDifference * HEIGHT_DIFFERENCE_FACTOR);
-          popupWindow.setHeight(height);
-        } else {
-          height = heightDifference + correctionFactor;
-          popupWindow.setHeight(height);
-        }
-        popupWindow.setWidth(rect.right);
+      Rect rect = Utils.windowVisibleDisplayFrame(context);
```
```suggestion
      final Rect rect = Utils.windowVisibleDisplayFrame(context);
```

emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
```diff
@@ -156,7 +151,8 @@
 
     popupWindow.setContentView(emojiView);
     popupWindow.setInputMethodMode(PopupWindow.INPUT_METHOD_NOT_NEEDED);
-    popupWindow.setBackgroundDrawable(new BitmapDrawable(context.getResources(), (Bitmap) null)); // To avoid borders and overdraw.
+    popupWindow.setBackgroundDrawable(
+        new BitmapDrawable(context.getResources(), (Bitmap) null)); // To avoid borders and overdraw.
```
vanniktech: nit: Can you revert this?

emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
```diff
@@ -176,35 +170,43 @@
...
-    // Manually dispatch the event. In some cases this does not work out of the box reliably.
-    context.getWindow().getDecorView().getViewTreeObserver().dispatchOnGlobalLayout();
+    if (Utils.shouldOverrideRegularCondition(context, editText)) {
+      editText.setImeOptions(editText.getImeOptions() | EditorInfo.IME_FLAG_NO_EXTRACT_UI);
+      inputMethodManager.restartInput(editText);
+    }
+
+    inputMethodManager.showSoftInput(editText, InputMethodManager.RESULT_UNCHANGED_SHOWN,
+        new ResultReceiver(null) {
+          @Override
```
vanniktech: nit: should be on it's own line

emoji/src/main/java/com/vanniktech/emoji/Utils.java
```diff
@@ -50,19 +51,37 @@
...
-    return dpToPx(context, context.getResources().getConfiguration().screenHeightDp);
+  static int getInputMethodHeight(Context context) {
+    try {
+      InputMethodManager imm =
+          (InputMethodManager) context.getApplicationContext().getSystemService(Context.INPUT_METHOD_SERVICE);
+      Class clazz = imm.getClass();
+      Method method = clazz.getDeclaredMethod("getInputMethodWindowVisibleHeight");
+      method.setAccessible(true);
+      return (int) (Integer) method.invoke(imm);
+    } catch (Exception e) {
```
vanniktech: Can we please log this?

mario:
All done.

mario:
All done!

mario:
This PR fixes #328 and #340.

rubengees:
Sorry for coming back so late! Just tried the latest changes and they seem to not work at all :/ Nothing happens for me when I click on the emoji button in the sample (real device Oneplus 5T, SwiftKey keyboard)...

I also get memory leaks related to the `ResultReceiver`.

mario:
@rubengees can you check what does Utils.getInputMethodHeight returns for you? In the mean time I'll take a look at the receiver.

mario:
Receiver leak should now be fixed.

mario:
You keep disappearing :P I found the method I call via reflection doesn't exist everywhere. Trying to workaround now.

rubengees:
> Can you check what does Utils.getInputMethodHeight returns for you?

590.

Some logic in `updateKeyboardState` seems to be off, directly after opening, the `OnGlobalLayoutListener` seems to be called again and leads to `dismiss` being called.

mario:
@rubengees pushed a fix + fallback.

rubengees:
Works perfectly now!

mario:
Great!

mario:
Would you mind investigating the leak as I have a really awful headache atm?

I would guess its just a matter of setting it to null at the right place.

Formatting wise I think I fixed all the complaints.

mario:
Fixed the leak. There's another that I'm pretty sure is related to Android SDK, but yea... should be all nice now.

mario:
@vanniktech let me know if I need to do anything else to get this merged so we could have a release out in the next few days.

vanniktech:
@rubengees approval and a green CI build ;)

rubengees:
@mario I can't really follow the flow in the `EmojiPopup` anymore code wise, but it works great! I also tried it in a real applications and same results there.

But I still get the leak with the `ResultReceiver` :/

Leak

mario:
@vanniktech and @rubengees - could you point me to the places that need fixing for the build to go green? I seem to miss this somehow :-/

I'll take a look once again at the ResultReceiver, but my suspicion is this is related to Android SDK itself.

@rubengees happy to explain any parts of the code - sorry if it doesn't make sense, the keyboard handling is one of the worst parts of the Android SDK - EVER!

rubengees:
> Could you point me to the places that need fixing for the build to go green?

I usually run `./gradlew build` locally until it reports success. That is basically what CI does also. But good news: I pushed a commit that fixes the remaining style issues.

> I'll take a look once again at the ResultReceiver, but my suspicion is this is related to Android SDK itself.

Had a quick look at it, it seems to be not. It is not a static class and internally uses the `context` of the outer class. That is a leak, since the `context` can't be garbage collected.
I guess you are better able to fix this though, since you know the code better.

> Happy to explain any parts of the code

I would really like a high level explanation of how the code flow is from tapping the emoji button to showing the keyboard. Not required for merging of course, but would like to understand :)

> The keyboard handling is one of the worst parts of the Android SDK - EVER!

I wholeheartedly agree!

emoji/src/main/java/com/vanniktech/emoji/Utils.java
```diff
@@ -71,29 +61,33 @@ static boolean shouldOverrideRegularCondition(@NonNull final Activity context, f
     return false;
   }
 
-  static int getInputMethodHeight(final Context context, final View rootView) {
+  @SuppressWarnings({"unchecked", "JavaReflectionMemberAccess"}) static int getInputMethodHeight(final Context context, final View rootView) {
```
rubengees: @vanniktech I really dislike the "put annotations on the same line" rule, since it leads to awful things like this. Do we really want to enforce that?

mario:
Thanks for the style fixes - I really don't sleep much this year so I must have missed the proper part of the CI output :(

rubengees:
The CI doesn't output the findbugs issues (on which the build failed), which is why I recommend to run the build always locally.

mario:
Less trivial than I though, but I think I got it. @rubengees could you please test?

emoji/src/main/java/com/vanniktech/emoji/Utils.java
vanniktech: Yeah we should not enforce it.

mario:
There were two separate leaks that I've hopefully fixed correctly - and hopefully the build turns green this time.

mario:
Tadadadadada .... it's green! :D

mario:
Flow (Assuming we just started)

- Click the emoji toggle button
- Check if the EditText we're using has a noExtractUI flag and we're in a landscape orientation
- If we are, store the original IME options, set the flag on the editText and restart input so the changes would get picked up by the keyboard right away

We do this because the global layout doesn't get into effect with fullscreen EditText since the keyboard is in another process and in a different layout.

We set the receiver to know what happens when we try to show the keyboard, and toggle the keyboard. If a keyboard is already shown, it'll show the Emoji popup.

- Now that we basically have a regular keyboard and views, we can measure keyboard height using two alternatives:

- hidden method to get keyboard height
- magic calculations as a fallback in case reflection fails (even though the fallback also uses reflection to a degree)

- on dismiss, we basically have to do three things:
    - clear the emoji receiver so we don't leak
    - check if we changed the original imeOptions of the editText and if so, restore them
    - cancel autofillManager so we don't leak EmojiSpans

Let me know if I failed to explain something properly and I'll happily do it.

emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
```diff
@@ -229,12 +224,22 @@ public void dismiss() {
...
       final InputMethodManager inputMethodManager = (InputMethodManager) context.getSystemService(Context.INPUT_METHOD_SERVICE);
+
       if (inputMethodManager != null) {
         inputMethodManager.restartInput(editText);
       }
+
+      if (SDK_INT >= O) {
+        AutofillManager autofillManager = context.getSystemService(AutofillManager.class);
+        if (autofillManager != null) {
+          autofillManager.cancel();
```
rubengees: Great! I sometimes saw this leak in my app, but always thought that was an Android SDK one.

emoji/src/main/java/com/vanniktech/emoji/EmojiResultReceiver.java
```diff
@@ -0,0 +1,32 @@
+package com.vanniktech.emoji;
+
+import android.os.Bundle;
+import android.os.Handler;
+import android.os.ResultReceiver;
+
+public class EmojiResultReceiver extends ResultReceiver {
```
vanniktech:
```suggestion
public final class EmojiResultReceiver extends ResultReceiver {
```

emoji/src/main/java/com/vanniktech/emoji/EmojiResultReceiver.java
```diff
@@ -0,0 +1,32 @@
...
+  private Receiver receiver;
```
vanniktech:
```suggestion
  @Nullable private Receiver receiver;
```

emoji/src/main/java/com/vanniktech/emoji/Utils.java
```diff
@@ -50,19 +45,73 @@
...
-  static int getScreenHeight(@NonNull final Activity context) {
-    return dpToPx(context, context.getResources().getConfiguration().screenHeightDp);
+  @SuppressWarnings({"unchecked", "JavaReflectionMemberAccess"}) static int getInputMethodHeight(final Context context, final View rootView) {
+    try {
+      final InputMethodManager imm = (InputMethodManager) context.getApplicationContext().getSystemService(Context.INPUT_METHOD_SERVICE);
+      final Class inputMethodManagerClass = imm.getClass();
+      final Method visibleHeightMethod = inputMethodManagerClass.getDeclaredMethod("getInputMethodWindowVisibleHeight");
+      visibleHeightMethod.setAccessible(true);
+      return (int) visibleHeightMethod.invoke(imm);
+    } catch (NoSuchMethodException exception) {
```
vanniktech: nit: can be grouped together

emoji/src/main/java/com/vanniktech/emoji/Utils.java
```diff
@@ -50,19 +45,73 @@
...
+      attachInfoField.setAccessible(true);
+      final Object attachInfo = attachInfoField.get(rootView);
+      if (attachInfo != null) {
+        final Field stableInsetsField = attachInfo.getClass().getDeclaredField("mStableInsets");
+        stableInsetsField.setAccessible(true);
+        return ((Rect) stableInsetsField.get(attachInfo)).bottom;
+      }
+    } catch (NoSuchFieldException noSuchFieldException) {
+      Log.w(TAG, noSuchFieldException.getLocalizedMessage());
+    } catch (IllegalAccessException illegalAccessException) {
```
vanniktech: same

emoji/src/main/java/com/vanniktech/emoji/Utils.java
rubengees: No they can't, tried that and the IDE complained, because this gets compiled to catch an exception only available on Java 8 if I recall correctly. Catching generic `Exception` should also not be done since it hides `RuntimeException`s on which we want to crash.

emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
mario: :) It's almost like magic! :)

mario:
@rubengees can we merge? After that, @vanniktech can make a release.

emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
mario: I'm pretty sure this is the only available open source lib these days that handles keyboard absolutely correctly btw, which is a great accomplishment - thanks to both @rubengees and @vanniktech for pushing when things didn't work as they were supposed to!

rubengees:
I'll make a last test soon, then I'll approve, thanks for the explanation!

vanniktech:
Thanks @mario really appreciate the effort you put into this! Feel free to merge!

mario:
Please make a release now, thanks! :)

vanniktech:
Will do first thing in the morning tomorrow

PR #336, opened 2019-01-27
Show duplicate emojis in the input but not in the picker

rubengees:
The generator is modified to add duplicate emojis to the data structures, but with a new flag `isDuplicate`. The `EmojiArrayAdapter` uses this information to filter duplicates from the picker.

Pasting the ["Person Facepalming" emoji from Emojipedia](https://emojipedia.org/face-palm/) (Codepoint: U+1F926) works now as expected.

~I have not regenerated yet to keep the change set small, so CI will obviously fail.~


generator/.gitignore
```diff
@@ -66,6 +66,3 @@ com_crashlytics_export_strings.xml
 crashlytics.properties
 crashlytics-build.properties
 fabric.properties
-
-yarn.lock
```
rubengees: This file should actually be commited, so I added it while I was at it.

generator/package.json
```diff
@@ -10,11 +10,11 @@
...
-    "emoji-datasource": "^4.0.4",
-    "emoji-datasource-apple": "^4.0.4",
-    "emoji-datasource-emojione": "^4.0.4",
-    "emoji-datasource-google": "^4.0.4",
-    "emoji-datasource-twitter": "^4.0.4",
+    "emoji-datasource": "~4.0.4",
+    "emoji-datasource-apple": "~4.0.4",
+    "emoji-datasource-emojione": "~4.0.4",
+    "emoji-datasource-google": "~4.0.4",
+    "emoji-datasource-twitter": "~4.0.4",
```
rubengees: I have set the version of our data set to the last one supporting EmojiOne. To improve backwards compability, I think we should keep it in for 0.6.0 and remove it for the version following that.

generator/index.js
```diff
@@ -35,10 +35,10 @@ const targets = [{
 }];
 
 /**
- * Emoji codepoints to globally ignore.
+ * Emoji codepoints which are duplicates. These are marked as such in the generated code.
  * @type {string[]}
  */
-const ignore = ["1F926", "1F937", "1F938", "1F93C", "1F93D", "1F93E", "1F939"];
+const duplicates = ["1F926", "1F937", "1F938", "1F93C", "1F93D", "1F93E", "1F939"];
```
mario: Do the duplicates have different information in the original datasource or what? Is it worth filling bugs at the datasource repo?

generator/package.json
mario: 👍

emoji/src/main/java/com/vanniktech/emoji/Utils.java
```diff
@@ -79,6 +83,19 @@ static void backspace(@NonNull final EditText editText) {
     editText.dispatchKeyEvent(event);
   }
 
+  static List<Emoji> filterDuplicates(List<Emoji> emojis) {
+    final List<Emoji> result = new ArrayList<>(emojis);
```
vanniktech: Can we reduce the number of allocations that we are doing here? I don't think we can looking at it from GitHub but I'd like to :D

emoji/src/main/java/com/vanniktech/emoji/emoji/Emoji.java
```diff
@@ -19,24 +19,29 @@
 
   @NonNull private final String unicode;
   @DrawableRes private final int resource;
+  private final boolean isDuplicate;
```
vanniktech: Can we store a reference to the one that is replacing this Emoji instead? That way we could trace them back and have a nice linked structure.

emoji/src/main/java/com/vanniktech/emoji/Utils.java
rubengees: Yeah, I thought about that while doing this too, but haven't had an idea yet.

emoji/src/main/java/com/vanniktech/emoji/emoji/Emoji.java
rubengees: A reference to the object is hard (since these are added to the array at the same time, so we habe no instance of the other emoji, while instantiating this one), but the unicode would be possible. Would that suffice?

mario: I'd be happy with that.

emoji/src/main/java/com/vanniktech/emoji/Utils.java
mario: Um, can't we just convert the list to a set?

generator/index.js
mario: Thanks for the explanation.

emoji/src/main/java/com/vanniktech/emoji/Utils.java
rubengees: We can't, the duplicates are not the same equals-wise. Even if we could, I don't know if that would be a good idea performance wise, since we would have to use `LinkedHasSet` (we need ordering) and that would be quite some overhead for little gain.

emoji/src/main/java/com/vanniktech/emoji/emoji/Emoji.java
rubengees: I just had the thought that we *may* be able to abuse the variant mechanism for that. Variants are not shown per default in the picker and we could filter the variants to show in the variant popup by the same mechanism, but way more performant since that are only 6ish elements at most.

Semantically it is a bit off, but duplicates could still be considered "variants" I think.

What do you think?

mario: Not a bad idea at all!

vanniktech: Can't we first construct faceplam emoji 1? And then make emoji 2 contain a reference to emoji1 since it's duplicated?

I would not misuse the `variants`. That's a different principle and we should mix them up.

rubengees: Possible, but hard. We'd need quite some additional logic in the generator and the templates for that.

vanniktech: Then let's leave it as is with the boolean

rubengees: Let's to that then. Might come back to it in a follow up.

vanniktech:
Let's get this into 0.6.0 as well.

mario:
@vanniktech awesome, that + #329 once @rubengees gets the time to check it out will make for an awesome release.

vanniktech:
Plus if you have the time, also the licensing files.

mario:
@vanniktech is this ready for merge?

vanniktech:
No. Travis is falling since he didn't generate everything to make it easier to review ;)

rubengees:
I found a way to reduce the allocations in the `filterDuplicates` method by doing that and `asList` in one step.

I believe this is ready to merge then, once I regenerate and CI passes.

vanniktech:
```
/home/travis/build/[secure]/Emoji/emoji-google/src/main/java/com/[secure]/emoji/google/GoogleEmoji.java:38: error: incompatible types: int[] cannot be converted to int
    super(codePoints, -1);
          ^
/home/travis/build/[secure]/Emoji/emoji-google/src/main/java/com/[secure]/emoji/google/GoogleEmoji.java:45: error: incompatible types: int cannot be converted to int[]
    super(codePoint, -1);
          ^
/home/travis/build/[secure]/Emoji/emoji-google/src/main/java/com/[secure]/emoji/google/GoogleEmoji.java:52: error: no suitable constructor found for Emoji(int,int,Emoji[])
    super(codePoint, -1, variants);
    ^
```

rubengees:
Yeah, that is fixed after regenerating. I'll go ahead and do that then.

rubengees:
I forgot to fix the tests, which I now did in the separate commit d006bd2 (plus a new test for the filter method).

I also regenerated and when CI passes, we can go ahead and merge :slightly_smiling_face:

vanniktech:
Thanks @rubengees. What should we do with EmojiOne? Remove our provider?

rubengees:
Yes, we need to remove it from the generator, the README and remove the already existing classes. Will put up a PR for that shortly :slightly_smiling_face:

PR #338, opened 2019-02-03
Remove EmojiOne

rubengees:
This removes all traces of EmojiOne from the code base. The EmojiOne team does not seem to want their work to be distributed as seen [here](https://github.com/iamcal/emoji-data/commit/a31b052284fd473381048c2aeba456404ec5457a). 

As discussed before, this should only be merged, *after* 0.6.0 has been released to improve backwards compatibility.


vanniktech:
@rubengees this is no longer blocked

rubengees:
I have merged master and this is ready to merge then.

PR #353, opened 2019-03-08
Change default Keyboard + ViewPager animation.

### LINKED ISSUE #347: Change default animation in demo

vanniktech:
Fixes #347


mario:
Hey, so I don't understand why this was merged as-is without waiting at least a day :(

- why did you remove the slide animations? Even if we don't use them, they could have served as an example / someone might have used them
- their mention was not removed from the README
- if you removed the view pager animation lib, you can also probably disable the jetifier as that was the only thing holding it down

vanniktech:
Whops. You're totally right. Will wait for your input next time.

> * why did you remove the slide animations? Even if we don't use them, they could have served as an example / someone might have used them

Correct me if I'm wrong but there were two animations. One was always unused. So I went with the fade in / fade out. 

> * their mention was not removed from the README

I don't quite follow.

> * if you removed the view pager animation lib, you can also probably disable the jetifier as that was the only thing holding it down

I thought so too but fastlane is still missing AndroidX :(

PR #365, opened 2019-05-30
Make EmojiManager's initialization methods synchronized

rubengees:
These methods are all operating on the static `INSTANCE` field and are modifying it, creating race conditions. While in practice this should never happen (`install` should be called synchronously in `onCreate`), I have a specific use case which requires synchronization: Async installation. If not synchronized, a call could happen directly after `install`, throwing an Exception, because the installation is not done yet.


rubengees:
Sorry, for the noise @vanniktech, I had problems getting the static code analysis tools to accept these changes 😁

mario:
Looks decent :D

Issue #366, opened 2019-06-05
Compress the PNG's


rocboronat:
This issue is just to share my results compressing the iOS png's with ImageOptim, an open source application to compress images.

I don't know if the images that are being released are already compressed, but just for the information: here's a screenshot of the results I had.

<img width="882" alt="compressed-images" src="https://user-images.githubusercontent.com/456499/58932095-dfd0fb00-8727-11e9-8ff9-390d58617a16.png">

Feel free to close this issue as soon as you thing it's not useful :·)

Thanks for the library, by the way! :D

rubengees:
We actually compress the images already with [PngQuant](https://pngquant.org/) and [Zopfli](https://github.com/google/zopfli) (automated in our [generator](https://github.com/vanniktech/Emoji/tree/master/generator).

As you can see, the savings for the actually large relevant file *emoji_ios_sheet.png* is only 0.2% since it is already heavily compressed. The category images are currently not compressed if I remember correctly, so that could be improved :slightly_smiling_face:.

rocboronat:
Great! Good work automating it!

Do you want me to compress the category images and launch a PR?

rubengees:
Sure, please do!

rocboronat:
That's all we can do related to this issue. Let's close it! Thanks for your quick reply, @rubengees :·)

PR #367, opened 2019-06-05
Optimise the category png's to save some space

rocboronat:
This PR is related to the conversation we had on #366.

So, the category images have been optimised with ImageOptim, and it saved about 31% per file on average. That's about 4KB per project! Wow. Huge change.


rocboronat:
Wow, according to GitHub, the compression is between 65% and 75%. That's even more! 😂

rocboronat:
@vanniktech done! I totally missed that pictures 👍

rocboronat:
🎉

vanniktech:
Thanks for your contribution!

rocboronat:
My pleasure. It's not a big one, but I was wondering which files I could strip from my app, and just thought what could happen if I optimise this lib's pics. Thanks for your work @vanniktech @rubengees 😄

PR #373, opened 2019-07-30
Allow popup height to be changed with a setter

VitalyKuznetsov:
Adding functionality to EmojiPopup to be opened in a different mode:
Keyboard height is not dynamically calculated, but set manually instead.


mario:
I'd prefer the approach that was mentioned earlier - allow setting height in the builder and if it's 0, calculate the height - otherwise use set height.

VitalyKuznetsov:
Consider the following case:

1. User opens app and uses emojiPopup -> height is set in builder.
2. User changes keyboard height in settings.
3. Goes back to the app, emojiPopup has not been destroyed yet.
4. Upon keyboard opening we save its height, however emojiPopup height cannot be changed, until a new instance is created.

For such scenario a simple setter would be quite useful.

VitalyKuznetsov:
So what do you think?

VitalyKuznetsov:
Hey, @mario , take a look!
Maybe it is more like what you are looking for?

emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
```diff
@@ -398,6 +432,9 @@ private Builder(final View rootView) {
       emojiPopup.onEmojiPopupShownListener = onEmojiPopupShownListener;
       emojiPopup.onEmojiPopupDismissListener = onEmojiPopupDismissListener;
       emojiPopup.onEmojiBackspaceClickListener = onEmojiBackspaceClickListener;
+      if (this.popupWindowHeight > 0) {
```
```suggestion
      if (popupWindowHeight > 0) {
```

VitalyKuznetsov:
Yeah sure, but the gboard has an option to change its height from settings, which may occure when emojipopup is already created.
In this case I'd like to change its height without creating a new instance.

mario:
I'll test this in the coming days.

VitalyKuznetsov:
Any updates? @mario

mario:
Looks ok.

vanniktech:
@VitalyKuznetsov can you fix the CI?

mario:
Can be merged after rebase.

vanniktech:
@VitalyKuznetsov are you still using this method and if so; can you show me your code which detects that the keyboard height has been changed and hence you are manually calling this method?

PR #377, opened 2019-09-10
Ship OnlyEmojisInputFilter & MaximalNumberOfEmojisInputFilter.

vanniktech:


mario:
Looks decent.

emoji/src/main/java/com/vanniktech/emoji/OnlyEmojisInputFilter.java
```diff
@@ -0,0 +1,33 @@
...
+
+import android.text.InputFilter;
+import android.text.Spanned;
+import androidx.annotation.Nullable;
+
+public final class OnlyEmojisInputFilter implements InputFilter {
+  @Nullable private final Integer maxCount;
+
+  public OnlyEmojisInputFilter() {
+    maxCount = null;
```
rubengees: Isn't this a constraint which belongs into another Filter and maybe not even in the library?

vanniktech: Dunno. :man_shrugging: I feel like they can kinda live in the same InputFilter.

rubengees: I have no strong feeling about this, but these are two separate concerns.
Also, ìsn't it already possible to set the maxLength with `android:maxLength`?

vanniktech: You should know that Emojis have different lengths. This won't work. I'll try splitting them up - if that works it's even better.

vanniktech: I like it even more now. Thanks @rubengees 
Composition is a nice thing

rubengees: Ah you are right, completly forgot about emojis having multiple chars :man_facepalming:!

vanniktech:
Waiting for your review @mario before merging

vanniktech:
So what now @mario?

mario:
Merged1

PR #378, opened 2019-09-10
Support use case where only Emoji Dialog should be shown.

vanniktech:


mario:
Ooook.

emoji/src/main/java/com/vanniktech/emoji/EmojiEditText.java
```diff
@@ -86,4 +87,25 @@ public final void setEmojiSizeRes(@DimenRes final int res) {
   public final void setEmojiSizeRes(@DimenRes final int res, final boolean shouldInvalidate) {
     setEmojiSize(getResources().getDimensionPixelSize(res), shouldInvalidate);
   }
+
+  /** Forces that only Emojis are allowed as input. */
+  public void forceEmojisOnly(final EmojiPopup emojiPopup) {
+    setOnFocusChangeListener(new ForceEmojisOnlyFocusChangeListener(emojiPopup));
```
rubengees: This doesn't do what the method name suggests, it shows the popup once the EditText is focused instead.

emoji/src/main/java/com/vanniktech/emoji/OnlyEmojisInputFilter.java
```diff
@@ -0,0 +1,33 @@
+package com.vanniktech.emoji;
+
+import android.text.InputFilter;
+import android.text.Spanned;
+import androidx.annotation.Nullable;
+
+public final class OnlyEmojisInputFilter implements InputFilter {
```
rubengees: This is also present in the other PR, did you mix these up?

vanniktech: Seems like it - yeah.

emoji/src/main/java/com/vanniktech/emoji/EmojiEditText.java
vanniktech: So `showsEmojisOnly`. With this approach, there's no way to get to the normal keyboard though and insert normal characters.

rubengees: Ah no I get what this is supposed to do! I got confused since I thought this method should control if the `OnlyEmojisInputFilter` should be applied or not, but it's somethings else.

Maybe a name like `disableKeyboardInput` or something would be more clear?

rubengees: If we add this, we may want a method to revert this behavior e.g. remove the `OnFocusListener`.

vanniktech:
This is ready now @mario & @rubengees

vanniktech:
Waiting for @mario's review.

PR #380, opened 2019-09-19
Don't use bundled AppCompat emojis. Instead download the font.

vanniktech:
Much much better. Thanks to @stefanhaustein for providing the integration with Google's AppCompat.


mario:
Cool, but where are these certs from?

vanniktech:
Google sample's everyone is using 'em.

mario:
Generally saying I'm more for bundled option (privacy related), but oh well. Looks good otherwise.

vanniktech:
It's the sample app and everyone is showing how to use the bundled one so I'm for the font one. In general, people can choose what they want since we support both options.

PR #383, opened 2019-09-24
Support textAllCaps option. Fixes #361

### LINKED ISSUE #361: EmojiTextView has no effect when set to all caps
- Version of the library: com.vanniktech:emoji-twitter:0.6.0
- Device version: Android 4.4.2

When I set `android:textAllCaps="true"` on an `EmojiTextView`, it reverts back to showing the default stock emojis rather than showing the emojis provided by the library. Here is a sample layout.

```
<!-- The dark background and large text size are just to make the emojis clearer to see -->
<LinearLayout
    xmlns:android="http://schemas.android.com/apk/res/android"
    android:layout_width="match_parent"
    android:layout_height="match_parent"
    android:orientation="vertical"
    android:padding="16dp"
    android:background="#444">

    <!-- what it should be -->
    <com.vanniktech.emoji.EmojiTextView
        android:id="@+id/text1"
        android:layout_width="wrap_content"
        android:textSize="30sp" />

    <!-- what it is -->
    <com.vanniktech.emoji.EmojiTextView
        android:id="@+id/text2"
        android:layout_width="wrap_content"
        android:layout_marginTop="16dp"
        android:textSize="30sp"
        android:textAllCaps="true" />

    <!-- what it shouldn't be -->
    <TextView
        android:id="@+id/text3"
        android:layout_width="wrap_content"
        android:layout_marginTop="16dp"
        android:textSize="30sp" />

</LinearLayout>
```
Here is the activity code (in Kotlin).

```
    // EmojiManager.install(TwitterEmojiProvider()) is called in the Application
    override fun onCreate(savedInstanceState: Bundle?) {
        super.onCreate(savedInstanceState)
        setContentView(R.layout.test_emojis)
        val emojisToTest = """🏅🇳🇬❤👩🏿‍⚕️"""
        text1.text = emojisToTest
        text2.text = emojisToTest
        text3.text = emojisToTest
    }
```

This is the result.

Screenshot_2019-05-13-07-20-20 1

The first one is the expected output with an `EmojiTextView`. The third is the expected output with a regular `TextView`. The second one is also an `EmojiTextView`, but because of `android:textAllCaps="true", it uses the default emojis, which are not always correct. 

In this sample, for example, the "black doctor" emoji is a combination of two emojis and a skin color modifier, but instead the emojis are shown separately, and also there is no "sports medal" emoji and the flag emoji shows as letters instead of, well, a flag. This is to be expected for a `TextView` because my device does not have those emojis by default, but that's what the library is for, so it shouldn't still be a problem with `EmojiTextView`.

Am I doing something wrong here, or is this a bug?

mario:

### DIFF
--- a/app/src/main/res/layout/item_adapter_chat.xml
+++ b/app/src/main/res/layout/item_adapter_chat.xml
@@ -11,6 +11,7 @@
       android:layout_width="wrap_content"
       android:layout_height="wrap_content"
       android:textIsSelectable="true"
+      android:textAllCaps="true"
       app:emojiSize="@dimen/emoji_size_default"
       tools:text="my text"
       />
--- a/app/src/main/res/layout/item_adapter_chat.xml
+++ b/app/src/main/res/layout/item_adapter_chat.xml
@@ -11,7 +11,6 @@
       android:layout_width="wrap_content"
       android:layout_height="wrap_content"
       android:textIsSelectable="true"
-      android:textAllCaps="true"
       app:emojiSize="@dimen/emoji_size_default"
       tools:text="my text"
       />
--- a/emoji/src/main/java/com/vanniktech/emoji/EmojiManager.java
+++ b/emoji/src/main/java/com/vanniktech/emoji/EmojiManager.java
@@ -133,8 +133,8 @@ import static com.vanniktech.emoji.Utils.checkNotNull;
       }
 
       final String regex = patternBuilder.deleteCharAt(patternBuilder.length() - 1).toString();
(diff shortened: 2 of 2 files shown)


app/src/main/res/layout/item_adapter_chat.xml
```diff
@@ -11,6 +11,7 @@
       android:layout_width="wrap_content"
       android:layout_height="wrap_content"
       android:textIsSelectable="true"
+      android:textAllCaps="true"
```
vanniktech:
```suggestion

```

Let's just skip the sample

PR #389, opened 2019-10-02
Fix keyboard calculation for API v21+ by introducing start/stop into EmojiPopup

mario:
Also fixes the issue with the dialog thing.


mario:
Makes it way way smoother.

mario:
Please test this as much as you can :) Thanks!

mario:
Lint complains of simple things:

[ant:checkstyle] [ERROR] /home/travis/build/[secure]/Emoji/app/src/main/java/com/[secure]/emoji/sample/MainDialog.java:50: Line has trailing spaces. [RegexpSingleline]
1219
1220

Please fix this if that's all that's blocking this PR from being merged.

emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
```diff
@@ -214,7 +214,7 @@ void updateKeyboardStateOpened(final int keyboardHeight) {
     final int properWidth = Utils.getProperWidth(context);
 
     if (popupWindow.getWidth() != properWidth) {
-      popupWindow.setWidth(properWidth);E
```
mario: What was this o.O

vanniktech:
@mario I have no idea how that `E` ended up being there.

rubengees:
I have managed to produce weird behaviour:

- Open Dialog.
- Click Emoji button.
- Close dialog.
- Click Emoji button (normal one).
  - _Emoji popup does not appear._
- Click again.
  - _Weird behaviour (see screenshot)._

device-2019-10-02-185403

Tested on a Pixel 2 API 29 emulator.

mario:
OK @rubengees I know what the issue is. Now on to finding a way to fix it.

vanniktech:
:rocket:

mario:
Please check now. I've identified an issue with emojis only filter (not sure if it was there before) so I'll look into it in a follow-up PR.

app/src/main/java/com/vanniktech/emoji/sample/MainActivityMultiAutoCompeteTextView.java
```diff
@@ -49,7 +49,10 @@
     emojiButton.setColorFilter(ContextCompat.getColor(this, R.color.emoji_icons), PorterDuff.Mode.SRC_IN);
     sendButton.setColorFilter(ContextCompat.getColor(this, R.color.emoji_icons), PorterDuff.Mode.SRC_IN);
 
-    emojiButton.setOnClickListener(ignore -> emojiPopup.toggle());
+    emojiButton.setOnClickListener(ignore -> {
+      emojiPopup.start();
+      emojiPopup.toggle();
```
vanniktech: Let's call `start()` inside `toggle()`

mario:
Done.

vanniktech:
1. Click on the main edittext in the sample
2. Switch to the emoji keyboard by pressing the left icon
3. Press the back button 
4. The keyboard does get dismissed but the Emoji Dialog is still shown

mario:
Fixed.

README.md
```diff
@@ -98,6 +98,18 @@ emojiPopup.isShowing(); // Returns true when Popup is showing.
...
+```java
+emojiPopup.start()
+```
+
+The following code needs to be executed in onStop():
+
+```java
+emojiPopup.stop()
+```
+
```
rubengees: Can we reuse the `OnAttachStateChangeListener` or add [`ActivityLifecycleCallbacks`](https://developer.android.com/reference/android/app/Application.ActivityLifecycleCallbacks) (we already have an `Activity` context in the popup) to do this on our own to make this easier to use?

mario: We might be able to :P Can you do it? :D

mario:
Thanks for testing @rubengees !

mario:
Btw. in general there's a lots of potential for re-use when using multiple instances of Emoji magic in the same activity - but it's fairly complicated as I need to look where things are referenced and replace them at the proper time.

vanniktech:
I'd also suggest to leave things as they are. (With start/stop needing to be called explicitly)

mario:
Well personally I'd rather remove them, but let's just say I'd rather leave this for later than hassle with it now. 

Btw @vanniktech sent you an email yesterday.

vanniktech:
Agreed and I already replied ;)

rubengees:
I would propose to do this before the next release though, because that makes the library easier to use (and we will get less bug reports of people forgetting to call these methods 😀).

README.md
rubengees: I'll try 🙂

mario:
Thanks folks!

PR #391, opened 2019-10-02
Add emoji toolbar

mario:

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -92,6 +92,8 @@ ext {
   supportCardView = "androidx.cardview:cardview:1.0.0"
   supportAnnotations = "androidx.annotation:annotation:1.0.0"
   supportEmoji = "androidx.emoji:emoji-appcompat:1.0.0"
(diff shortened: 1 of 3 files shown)


vanniktech:
Maybe the solution is to ship a LayoutFactorty that replaces all TextView's by our TextView's. That would handle all cases with custom views etc.

mario:
@vanniktech my use case was simple so this is what I did. The long term solution is different of course :)

vanniktech:
Let me think about this a bit more

mario:
I think a factory is actually a nice idea.

rubengees:
If we do it like this, it should be a module on it's own, because it pulls in `material-components` which is large.

vanniktech:
> If we do it like this, it should be a module on it's own, because it pulls in `material-components` which is large.

Yup that's for sure!

I'm more inclined to provide a custom LayoutFactory though and then it'll just work _everywhere_

PR #392, opened 2019-10-02
Fix custom keyboard height

mario:


mario:
I still don't see a use case for this, but at least it now works as expected.

mario:
Requires previous PR to be merged.

mario:
Rebased.

vanniktech:
@mario needs a rebase :/

mario:
Ask @vanniktech :P

mario:
Ready for review.

emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
```diff
@@ -397,9 +401,11 @@ private Builder(final View rootView) {
      * If height is not 0 then this value will be used later on. If it is 0 then the keyboard height will
```
rubengees: These docs are not correct anymore,should be something like "if height is greater than 0".

mario: Fixed.

vanniktech:
Any opinion @rubengees ?

mario:
Poking @rubengees once more :)

mario:
Thanks folks!

PR #393, opened 2019-10-03
Fix emoji only filter

mario:
Previously, when you enabled emoji-only filter and focused the edittext the emoji popup would appear.

In the future, when you tried focusing it for the second time only a keyboard would open as in reality edittext is already focused. 

We now clear the focus after dismissing the emoji popup in case keyboard input is disabled.


mario:
@vanniktech mind fixing the warnings here?

vanniktech:
@mario I can do that once you're rebased

mario:
@rubengees fancy fixing warnings? My head is a mess today.

rubengees:
@mario Sure, done.

rubengees:
I found a way to break out of the emoji picker and show a normal keyboard.

Steps:

- Open keyboard.
- Open input method dialog (for me on Android 10 in the bottom right).
- Select other keyboard.

rubengees:
PMD reports the `EmojiPopup` as a possible god class now due to it's size, that might be harder to fix...

vanniktech:
Ignore it

emoji/build.gradle
```diff
@@ -23,6 +23,7 @@ dependencies {
   testImplementation rootProject.ext.testing.assertJ
   testImplementation rootProject.ext.testing.robolectric
   testImplementation rootProject.ext.testing.privateConstructor
+  androidTestAnnotationProcessor deps.support.test.autoService
```
vanniktech: What do we need this for?

app/build.gradle
```diff
@@ -68,6 +68,7 @@ dependencies {
   androidTestImplementation deps.support.test.runner
   androidTestImplementation deps.support.test.rules
   androidTestImplementation deps.fastLaneScreenGrab
+  androidTestAnnotationProcessor deps.support.test.autoService
```
vanniktech: same

emoji/build.gradle
mario: One of the lints complained about it missing.

mario:
> I found a way to break out of the emoji picker and show a normal keyboard.
> 
> Steps:
> 
> * Open keyboard.
> * Open input method dialog (for me on Android 10 in the bottom right).
> * Select other keyboard.

@vanniktech did you test this? Since I don't even see the input method dialog?

vanniktech:
> did you test this? Since I don't even see the input method dialog?

Not yet

vanniktech:
I found one way:

1. Check the checkbox
2. Click on the input field
3. Click on the dialog button
4. Dismiss the dialog
5. Click on the input field (Normal keyboard is shown instead of the emoji one)

mario:
Should be ok now @vanniktech and @rubengees

vanniktech:
@rubengees @mario how should we proceed with PR's so that they don't stay open that long?

I propose:

- At least one review
- Waiting for a second review for up to 3 days

mario:
Well in general I'm highly in favour of at least two reviews, but I guess we need a policy on how long we wait. I feel like 3 days is too little? How does 7 days work for you two?

vanniktech:
Sure. I'm fine with 7 days!

vanniktech:
I'll merge this now as discussed with @mario.

rubengees:
Sorry for being away so long, I agree regarding the proposed time for waiting.

vanniktech:
No need to be sorry. I hope everything is fine @ruben

rubengees:
Yes sure, was just a bit lazy in the last time 😄

PR #397, opened 2019-10-03
Automatically call start and stop when attaching/detaching EmojiPopup

rubengees:
As proposed in https://github.com/vanniktech/Emoji/pull/389, this changes `EmojiPopup` to automatically call it's `start` and `stop` methods, making the library easier to use.

This is done by calling the methods as part of the `OnAttachStateChangeListener` lifecycle.

Tested on my real device (API 29) and it looks like it just works™️.

### DIFF
--- a/emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
+++ b/emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
@@ -165,7 +165,7 @@ public final class EmojiPopup implements EmojiResultReceiver.Receiver {
   }
 
   /** Call this method in your #onStart method. */
-  public void start() {
+  void start() {
     if (SDK_INT >= LOLLIPOP) {
       context.getWindow().getDecorView().setOnApplyWindowInsetsListener(new View.OnApplyWindowInsetsListener() {
         int previousOffset;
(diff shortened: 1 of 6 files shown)


mario:
Start and stop don't need to be public then? :)

emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
```diff
@@ -165,7 +165,7 @@ void updateKeyboardState() {
   }
 
   /** Call this method in your #onStart method. */
```
mario: Hm we don't need this comment anymore.

emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
```diff
@@ -199,7 +199,7 @@ public void start() {
   }
 
   /** Call this method in your #onStop method. */
```
mario: Also don't need this.

emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
```suggestion

```

emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
```suggestion

```

vanniktech:
Thanks @rubengees I like it a lot more!

mario:
Merge

PR #399, opened 2019-10-03
RecentEmojiManager: Increase max emoji count to 100.

vanniktech:
40 might be too low.


mario:
Might be too low compared to what? XD

vanniktech:
:100:

vanniktech:
Sure we can make it configurable. I'd still like a default of 100

mario:
Ok, make it configurable with max 100 and then I'll approve this :P

rubengees:
I looked at other emoji pickers and all seem to have around 40 in their recent grids. Why do you plan to increase that? I wouldn't have anything against configuration though.

Telegram: 40
WhatsApp: 36
Slack: 21

mario:
Though I guess it's already possible to do that by providing your own RecentEmojiManager?

vanniktech:
Let's leave at 40 then

PR #402, opened 2019-10-14
Fix set popup window height method

mario:


mario:
Sorry @rubengees, I missed something here so updated now.

PR #420, opened 2019-11-18
Add animation support

BennieBe:
I have added methods to allow for either returning a bitmap or an animated bitmap if isAnimated boolean is true. I created a new Category with the Flag symbol and used one of the sprite sheets already existing in order to do the testing (in the app itself the chosen category is the last in the line of category icons, you can see two flags at the end, it is the last flag). I was planning to use the same method that the current long sprite sheets use to display all images, just that instead of showing all these bitmaps separately, I combine all bitmaps into one animation to be played when selected. You can find this in the loadAnimatedStrip method. In order to create an animation I use the existing loadStrip method to create a sequence of bitmaps using AnimationDrawable object. The animation played for me when I tried starting the method in loadAnimatedStrip directly, however, it does not animate when I try to start the animation in drawAnimation method. This is where I get stuck and I cannot find why it does not work. I would appreciate if someone can have a short look and give me some hints on what might be going wrong! ;)

### DIFF
--- a/emoji-ios/src/main/java/com/vanniktech/emoji/ios/IosEmoji.java
+++ b/emoji-ios/src/main/java/com/vanniktech/emoji/ios/IosEmoji.java
@@ -102,8 +102,6 @@ public class IosEmoji extends Emoji {
     Log.d("TAG", "Still working in AnimatedDrawable");
 
     final Bitmap strip = loadStrip(context);
-    Log.d("TAG", "Still working after loading strip");
-    Log.d("Strip is: ", strip.toString());
     final Bitmap[] cut = loadAnimatedStrip(strip);
 
     AnimationDrawable animationDrawable = new AnimationDrawable();
@@ -112,8 +110,7 @@ public class IosEmoji extends Emoji {
     }
     animationDrawable.setOneShot(false);
 
-    System.out.println("Frames: " + animationDrawable.getNumberOfFrames());
-    System.out.println("Frames: " + animationDrawable.isRunning());
+    System.out.println("Frames in getAnimatedDrawable Method: " + animationDrawable.getNumberOfFrames());
 
     return animationDrawable;
   }
@@ -124,7 +121,7 @@ public class IosEmoji extends Emoji {
 
     for(line = 0; line * SPRITE_SIZE_INC_BORDER < bitmap.getHeight(); line++) {
       bitmapStripList.add(Bitmap.createBitmap(bitmap, 1, line * SPRITE_SIZE_INC_BORDER, SPRITE_SIZE, SPRITE_SIZE));
-      Log.d("TAG", bitmapStripList.get(bitmapStripList.size() - 1).toString());
+      Log.d("TAG", "Frame: " + line + " " +  bitmapStripList.get(bitmapStripList.size() - 1).toString());
     }
 
     Bitmap[] bitmapStripArray = new Bitmap[bitmapStripList.size()];
--- a/emoji-ios/src/main/java/com/vanniktech/emoji/ios/IosEmoji.java
+++ b/emoji-ios/src/main/java/com/vanniktech/emoji/ios/IosEmoji.java
@@ -4,15 +4,21 @@ import android.content.Context;
 import android.content.res.Resources;
 import android.graphics.Bitmap;
 import android.graphics.BitmapFactory;
+import android.graphics.drawable.AnimationDrawable;
 import android.graphics.drawable.BitmapDrawable;
 import android.graphics.drawable.Drawable;
 import androidx.annotation.NonNull;
+
+import android.util.Log;
 import android.util.LruCache;
+import android.widget.ImageView;
 
 import com.vanniktech.emoji.emoji.CacheKey;
 import com.vanniktech.emoji.emoji.Emoji;
 
 import java.lang.ref.SoftReference;
+import java.util.ArrayList;
+import java.util.List;
 
 public class IosEmoji extends Emoji {
   private static final int CACHE_SIZE = 100;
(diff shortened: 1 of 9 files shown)


emoji/src/main/java/com/vanniktech/emoji/EmojiSpan.java
```diff
@@ -3,8 +3,11 @@
...
 import android.graphics.Canvas;
 import android.graphics.Paint;
+import android.graphics.drawable.AnimationDrawable;
 import android.graphics.drawable.Drawable;
 import android.text.style.DynamicDrawableSpan;
+import android.util.Log;
+
 import com.vanniktech.emoji.emoji.Emoji;
 
 final class EmojiSpan extends DynamicDrawableSpan {
```
rubengees: Instead of extending this span, we want a special `Span` implementation for animations I think. Drawing the animation ourselfs on the `Canvas` is hard.

emoji-ios/src/main/java/com/vanniktech/emoji/ios/IosEmoji.java
```diff
@@ -58,24 +64,70 @@ public IosEmoji(final int codePoint, final int x, final int y, final boolean isD
...
+  }
+
+  private Drawable getDrawable(final Context context, CacheKey key) {
     final Bitmap strip = loadStrip(context);
     final Bitmap cut = Bitmap.createBitmap(strip, 1, y * SPRITE_SIZE_INC_BORDER + 1, SPRITE_SIZE, SPRITE_SIZE);
     BITMAP_CACHE.put(key, cut);
     return new BitmapDrawable(context.getResources(), cut);
   }
 
+  @Override @NonNull public AnimationDrawable getAnimatedDrawable(final Context context) {
```
rubengees: We could just reuse the `getDrawable` method and decide there if we want an animated one. This way no change to the existing interface is necessary (the boolean would also not be necessary). Calling code could check if the returned `Drawable` is an `AnimatedDrawable` to decide what to do.

emoji-ios/src/main/java/com/vanniktech/emoji/ios/IosEmoji.java
```diff
@@ -58,24 +64,70 @@ public IosEmoji(final int codePoint, final int x, final int y, final boolean isD
...
+      animationDrawable.addFrame(new BitmapDrawable(context.getResources(), bitmap), 400);
+    }
+    animationDrawable.setOneShot(false);
+
+    System.out.println("Frames in getAnimatedDrawable Method: " + animationDrawable.getNumberOfFrames());
+
+    return animationDrawable;
+  }
+
+  private Bitmap[] loadAnimatedStrip(Bitmap bitmap) {
```
rubengees: I think https://developer.android.com/reference/android/graphics/drawable/AnimatedVectorDrawable or some gif implementation is a more performant solution instead of loading each bitmap individually.

BennieBe: @rubengees I was thinking of using AnimatedImageDrawable: https://developer.android.com/reference/android/graphics/drawable/AnimatedImageDrawable
It allows for using GIF files directly using an ImageView. However, it requires API Version 28
which is fairly new, is this an option?

mario:
Sorry, can someone clarify to me what is this PR about and why do we need animation support?

(Granted I do want to add stickers/etc support, but this can't interfere with the Emojis code)

BennieBe:
@mario Well, you kind of said it yourself, gif animation stickers are certainly of interest to be supported. Since I am using this project and I wanted to add gif support for stickers etc. I was thinking of implementing it in here directly. 

About the emojis code, I plan to make it in a way that the emojis code is not being touched and simply add to the current implementation a custom implementation for gif animations which can be added individually on a manual basis.

rubengees:
How I envision this to be implemented is by extending the emoji core with minimal support for animated content (most likely by a custom `Emoji` class, that overrides `getDrawable`). We would then have a custom `Span` and extend the `EmojiTextView` + friends classes to handle these. If we need concrete `Drawable` implementations, these should be a module on their own.
The emoji core would decide which `Span` to create by checking the type of the `Emoji` object and/or the type of the returned `Drawable`.

Users would then implement their own category returning objects of the new `Emoji` class.

BennieBe:
Yes, the way @rubengees outlined sounds like a good way of implementation without touching the emojis core, so I was planning on performing only extensions of the core which do not interfere with the current implementation.

@rubengees you suggested using a gif implementation, so I found this here: https://developer.android.com/reference/android/graphics/drawable/AnimatedImageDrawable
It is a Drawable object for drawing animated images (like GIF). I see it as a good build in Android way to use GIFs, would this be alright to use since it requires API Version 28?

mario:
Sorry for the delay @BennieBe - are you still interested in working on this?

BennieBe:
@mario I currently have a project where this might be useful so yes I am still willing to work on this

BennieBe:
@mario Can you please tell me if it is alright to use AnimatedImageDrawable? https://developer.android.com/reference/android/graphics/drawable/AnimatedImageDrawable
It does bump the required api version to 28. Or is it needed to look for an alternative?

mario:
@BennieBe I would use AnimatedImageDrawable for 28+ and https://github.com/koral--/android-gif-drawable for 17-27.

This lib currently supports 15 and 16 as well, but I think it's safe to simply not have support for animations/gifs in case you're on such low API levels.

mario:
Like @rubengees and you agreed earlier, please try to keep existing interface changes to minimum where possible.

BennieBe:
@mario this looks good to me thanks, I will try to keep interface changes to a minimum

vanniktech:
If we require a third party lib we should create a new module.

BennieBe:
@vanniktech This means that I can open up a new project and integrate into the existing project, correct?

mario:
@BennieBe it's basically like we have: emoji-google, emoji-twitter, emoji-*. That's how you should do it. 
emoji-animated comes to mind :D or whatever.

vanniktech:
`emoji-animated` sounds good

vanniktech:
What's the progress here?

vanniktech:
I'm closing this issue due to inactivity. If you have any further input on the PR, don't hesitate to reopen this PR or post a new one.

PR #442, opened 2020-05-30
Add SingleEmojiTrait to force single (replaceable) on an EditText

vanniktech:


app/src/main/java/com/vanniktech/emoji/sample/CustomViewActivity.java
```diff
@@ -31,6 +32,7 @@
 
       emojiButton = findViewById(R.id.customViewButton);
       editText = findViewById(R.id.customViewEditText);
+      new SingleEmojiTrait(editText);
```
rubengees, line 35: I feel like this should be a static method, reads a bit weird. Something like this?

`SingleEmojiTrait.install(editText);`

PR #446, opened 2020-06-01
Add shortcodes to emojis

rubengees:
This adds a new field with the shortcode to the `Emoji` class. Accordingly the generator has been extended.

This does not implement any UI or herlper methods yet.


mario:
Isn't a shortcode supposed to be prefixed with ":" ?

vanniktech:
Also how does slack know that `ant` could also be `bug` when I'm typing `bu`?

Screenshot 2020-06-02 at 13 24 40

mario:
Slack is wrong because an ant is not a bug xD

rubengees:
@vanniktech You're right, there can actually be more (and our datasource has them)! I'll change it so we have a `List<String>` there.

@mario The ":" prefix is a markup convention which is used to detect emojis in text. It's not part of any standard I think.

vanniktech:
Thanks @rubengees 🙏

PR #447, opened 2020-06-01
Multiple Emoji Popups in one Activity.

vanniktech:
This one is funny.

Currently, EmojiPopup can more or less only be used for one EditText. Few ways to break things:

- 1st case
  - Click on the first edittext & insert emoji
  - Click on the second edittext & insert emoji
    - Expected: emoji in second edittext
    - Actual: it's in the first one
- 2nd case
  - Click on the first edittext & on the trigger button
    - Expected: emoji popup should be rendered correrctly
    - Actual: it's way off screen
    - Note: 2nd trigger button works though

UI screenshot for some context:

android_screenshot_1591041532

@rubengees & @mario any ideas how to fix this? 

I feel like an EmojiPopup should be able to switch its EditText so we only create the popup once and can activate it on different EditTexts.


vanniktech:
Decided to go with a Dialog instead.

Issue #218, opened 2017-11-13
Add additional tabs to keyboard


holysheep:
Is it possible to insert stickers tab in emoji keyboard?

vanniktech:
For sure.

vanniktech:
Just for the record you can already place your own categories with your own emojis by implementing your own Provider and delegating to a standard one for the first few tabs.

ThangPk:
@vanniktech When will be there a next release including this feature?  Thanks.

vanniktech:
Someone has to implement it first.

vanniktech:
Decided that this is slightly out of the main purpose - Emojis - of this library. Hence I'll close this. I believe that I've seen some solutions that offer both solutions (Emoij + Stickers) in one.

3llomi:
@vanniktech i am working on adding a 'Stickers' tab, if you are interested i can make a PR, otherwise i would make my own fork if you are okay with that

vanniktech:
I'd be interested in stickers; though they'd need to live in their own artifact. Would you be open for that idea?

3llomi:
@vanniktech Yeah,sure, so it would be like 'emoji-with-stickers' for example right?

vanniktech:
Just, `emoji-stickers`.

3llomi:
#453 @vanniktech  please check the PR out.

nAkhmedov:
It would be better to add ability extra tabs.

PR #453, opened 2020-07-07
Added a Stickers Tab

3llomi:
#218 

let me know your thoughts.


vanniktech:
First of all, thank you very much. Can you just update the sample app so that everyone can toy around with it?

3llomi:
i've added that, please take a look at it.

vanniktech:
I'll have a thorough look at this on the weekend!

vanniktech:
I ran the sample app and I can see the sticker but nothing happens when I click there. Is this expected? The demo should at least have some kind of demo stickers.

android_screenshot_1594539359

3llomi:
when you click on the Stickers tab it should show 4 stickers, however when you click on one of the tabs the listener should be triggered, in the Sample i just added a Log statement for now, do you want me to add the Sticker to the RecyclerView for example whenever the users clicks on one of the tabs?

vanniktech:
Yeah. Sample should show something.

3llomi:
i've done that, check it out.

vanniktech:
I'm having a super hard time clicking the stickers. Only after pressing 20 times on it the tab switched. Also, is there a reason you didn't comply to the coding standards here? Reviewing is hard because some of the files got completely reformatted.

3llomi:
Sorry about that, i didn't notice it since i was testing on an emulator. probably it has something to do with the new navigation gestures especially the bottom part, because i've tested it on a Nexus 5 and it works well, however on devices that has new Navigation gestures the the stickers/emoji buttons don't work all the time.
anyways  i will check that.

also can you please tell what 'coding standards' do you mean? have i changed something unintentionally?

vanniktech:
Yeah just look at the git diff. You went from 2 spaces to 4.

I was testing on a Pixel 4 XL, so yes that could be the case.

3llomi:
Okay so after some investigation i've found that the new Android 10 gestures is what causes the issue since the buttons are very close to the bottom navigation.

so what came to my mind is checking if the user has enabled Full screen gestures we would add some padding at the bottom, the problem is, it leaves a big white space which is not good and it's not that efficient.

however if i tried to close the Keyboard when the emojis show up using `InputMethodManager`

```java
.setOnEmojiPopupShownListener(() -> {
 InputMethodManager imeManager = (InputMethodManager) getApplicationContext().getSystemService(Context.INPUT_METHOD_SERVICE);
 imeManager.hideSoftInputFromWindow(editText.getWindowToken(),0);
 })
```

 it would just work fine.

emoji-showcase

so i think the problem is that the Emojis window takes the whole space leaving no space for this small bar for keyboard  closing/choosing input method. (take a look at the image below).

Android Emulator - Pixel_3_XL_API_29_2:5554 2020-08-07 20-20-30

and BTW the same problem applies on Emojis without implementing Stickers at all, so the very last Emoji at the bottom if you clicked on it at the bottom part it will not be clickable if the user has enabled Fullscreen gestures.

vanniktech:
Whatsapp has the same layout (without padding) and there it does not interfere at all. Somehow this is possible.

3llomi:
Yes, that's correct, but i believe that WhatsApp **hides** the keyboard when you click on the Emoji Button and that dismisses the bottom bar as shown in the GIF. 
so if we can hide the keyboard when the Emoji window shows up maybe that would be a good solution?

vanniktech:
> so if we can hide the keyboard when the Emoji window shows up maybe that would be a good solution?

You're right. As far as I know the entires structor of this lib relies on the fact that the keyboard is shown. Otherwise calculations won't be working correctly. This is something that can be tried in a separate branch though.

vanniktech:
What's the state?

Being honest, it's unlikely I'll accept that PR the current state. It's a bit buggy and also it has changed a lot lot of unnecessary files. I do appreciate all of the work you've put in - so if you want to continue with this and this is blocking you feel free to continue doing this in your fork. I'm not really eager / keen on stickers. If I can get a good implementation for free, I'll take it, otherwise less is better and this lib won't support it.

3llomi:
Sorry, but i just don't have enough time anymore to implement the logic of 'hiding'  the keyboard which it fixes the issue of the 'Stickers tab'.

i will close this for now, and maybe some time in future i will implement the 'keyboard hiding' logic and after that maybe create a new PR to implement a Stickers tab 👍 .

kevin4dhd:
Any advance?

PR #460, opened 2020-07-25
Add EmojiCheckbox.

vanniktech:
@rubengees @mario do you have any idea why the Emoji's won't show up when setting them via the xml text attribute?

They're simply missing

Screenshot 2020-07-25 at 09 03 13


steve-pollack:
I'll take a look.

rubengees:
Found it and pushed a fix. The handling of the size was incorrect (for all views), but not sure why this was only problematic for the check box.

vanniktech:
@rubengees the other components also had the same problem, I just found it in the CheckBox first

Issue #476, opened 2020-10-14
EmojiPopup.Builder.fromRootView() reads from the disk on the main thread


lukesleeman:
`EmojiPopup.Builder.fromRootView()` performs a read from the disk on the main thread.   This is caused by the builder constructing a `RecentEmojiManager` in its constructor:

https://github.com/vanniktech/Emoji/blob/6d33b3a68ed9b3102c8d574a7669d44d7da1a57b/emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java#L373

The `RecentEmojiManager` then creates a shared preferences instance, which then causes the main thread disk read.  In our case this causes problems as we are running with strict mode enabled, and we are building our own `RecentEmoji` instance, so we don't even use the one that gets made in the constructor.

There are a few ways we can address this issue.   I think the most straightforward approach would be to modify the `EmojiPopup.Builder` so that the `recentEmoji` field is `@Nullable` then construct a `RecentEmojiManager` inside the `build` method if `recentEmoji` is still null.  That way people can set their own emoji managers and manage the threading of any disk reads in whatever way suits them.

All up it seems like a pretty minor change.  @vanniktech would you be happy to accept a PR along these lines?  I could submit something.

vanniktech:
Yes sure, PR sounds good.

lukesleeman:
I've submitted a small PR which fixes the issue https://github.com/vanniktech/Emoji/pull/478

PR #478, opened 2020-10-26
Only construct RecentEmojiManager if one hasn’t already been set

lukesleeman:
This PR fixes issue #476 - Previously the `RecentEmojiManager` was being constructed when the builder was created.  This PR changes the builders `recentEmoji` field to @Nullable and performs the initialisation when build is called.  This way people can set their own emoji managers and manage the threading of any disk reads in whatever way suits them.

### DIFF
--- a/emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
+++ b/emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
@@ -364,13 +364,12 @@ import static com.vanniktech.emoji.Utils.checkNotNull;
     @Nullable OnEmojiBackspaceClickListener onEmojiBackspaceClickListener;
     @Nullable OnEmojiClickListener onEmojiClickListener;
     @Nullable OnEmojiPopupDismissListener onEmojiPopupDismissListener;
-    @NonNull RecentEmoji recentEmoji;
+    @Nullable RecentEmoji recentEmoji;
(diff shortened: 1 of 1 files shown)


emoji/src/main/java/com/vanniktech/emoji/EmojiPopup.java
```diff
@@ -482,6 +481,10 @@ private Builder(final View rootView) {
       EmojiManager.getInstance().verifyInstalled();
       checkNotNull(editText, "EditText can't be null");
 
+      if(recentEmoji == null){
```
vanniktech, line 484:
```suggestion
      if (recentEmoji == null) {
```

vanniktech (dismissed):

vanniktech (approved):

rubengees (approved):
👍

mario (approved):
Sure, but why don't we do the same with variantEmoji then?

vanniktech:
> Sure, but why don't we do the same with variantEmoji then?

Can do in a separate PR.

lukesleeman:
> Sure, but why don't we do the same with variantEmoji then?

The main reason for extracting this one out was the `RecentEmojiManager` does a disk read on the main thread when its constructed.  🤞🤞  I think the variant emoji is a bit more "lightweight" to construct

Issue #485, opened 2021-01-19
EmojiUtils.isOnlyEmojis cannot detect 🗯 ,  🗨,  🕳, ❤, ❣ .... 


tricksilver04:
Thank you for filing an issue. If this is a bug that you want to report, please take the time to provide some information:

- Version of the library:  0.7.0
- Affected devices: Any
- Affected versions: Any

To replicate issue :

EmojiUtils.isOnlyEmojis("🗯")
EmojiUtils.isOnlyEmojis("🗨")
EmojiUtils.isOnlyEmojis("🕳")
EmojiUtils.isOnlyEmojis("❤")
EmojiUtils.isOnlyEmojis("❣")

* this happens to all emojis with 2 codepoints like this:
`
new GoogleEmoji(new int[] { 0x1F5E8, 0xFE0F }, new String[]{"left_speech_bubble"}, 30, 26, false),
new GoogleEmoji(new int[] { 0x1F5EF, 0xFE0F }, new String[]{"right_anger_bubble"}, 30, 27, false),
`

Issue #484, opened 2020-12-17
Incorrect heart emoji detection in EmojiUtils


vellrya:
- Version of the library: 0.7.0
- Affected devices: OnePlus 3
- Affected versions: Android 8.0

----------------------------------------------------------------------

Hello, some strange behaviour:
EmojiUtils.emojisCount("❤") return 0. 
Heart symbol is also ignored in method EmojiUtils.emojis (empty list returned).

Thats all, thank you for time spend on maintaing library :)

vanniktech:
Dupe of #485

PR #486, opened 2021-01-19
added  property to Emoji class to be used for pattern matching purposes

johnearlelevado:
This is to fix issues #485 and #484. I added a new property for Emoji for pattern matching purposes.  The issue was that whenever the emoji has more than one code point, EmojiUtils.isOnlyEmojis() doesn't work.


john-rsvp (approved):

vanniktech (commented):
@rubengees is this a valid approach to fix the problem?

mario:
@vanniktech if it works and it's better than what we currently have, why not? :)

mario:
@rubengees poke here again :P

vanniktech:
@tricksilver04 I've added test for this: https://github.com/vanniktech/Emoji/pull/578

I tried your changes locally and unignored the test, however it's failing for the red emoji heart. Can you please have another look?

rubengees:
@vanniktech I debugged this a bit and have the following findings:

- Some emojis can optionally have a [variant selector](https://emojipedia.org/variation-selector-16/) which denotes the previous emoji to be rendered as an emoji instead of text. We have multiple emojis that can have this variant selector.
- Our datasource includes information about this. If the emoji can have a variant selector, the `non_qualified` field is not `null`. I tried to extend the generator to take this into account and made the tests pass except for the red heart exclamation.
- I saw that we already specifically handle the star emoji with a variant selector. Interestingly that emoji does not appear in our data source as such a case.
- The red heart exclamation seems to be either invalid or a special case. The code points are [10083, 65039, 65039] or `2763-FE0F-FE0F` in hex. `2763-FE0F` would be the heart exclamation with variant selector, but as we can see it has an additional variant selector! I'm not sure if multiple variant selectors are valid but e.g. browsers seem to ignore the second one.

I can think of two solutions:

- Strip all variant selectors from the text before processing it in either the utils or when rendering the images. We want to display everything as an emoji either way.
- Extend the generator to include this information in our code and handle it accordingly. Weird cases like the double variant selector would not work properly though I think.

vanniktech:
> I saw that we already specifically handle the star emoji with a variant selector. 

Yes if I remember correctly, we've had a report about this in the past and did a fix for it. (without thinking about all of the other emojis like described here)

---

Can we do a mix that we adjust the generator to include the information and just get rid of the double variant selector ourself?

vanniktech:
@rubengees I just saw this

https://github.com/vanniktech/Emoji/blob/b1155d5146237d7c6e60734734a861eea10ce645/generator/index.js#L248

This is where we handle the star

PR #493, opened 2021-04-19
EmojiEditText cursor height is too small by wirting emojis followed by text fixes #492

### LINKED ISSUE #492: EmojiEditText cursor height is too small by wirting emojis followed by text
If you write an emoji followed by a text. The height of the emoji cursor is too small and the textfield height reduce. If you write text the textfield increase. Because the line height is bigger. Solution: It should be the same line height. 

- Version of the library: 0.7.0
- Affected devices: all
- Affected versions: 0.7.0

----------------------------------------------------------------------

denjoo:
If you write an emoji followed by a text. The height of the emoji cursor is too small and the textfield height reduce. If you write text the textfield increase. Because the line height is bigger. Solution: It should be the same line height.

### DIFF
--- a/emoji/src/main/java/com/vanniktech/emoji/EmojiSpan.java
+++ b/emoji/src/main/java/com/vanniktech/emoji/EmojiSpan.java
@@ -26,67 +26,68 @@ import android.text.style.DynamicDrawableSpan;
 import com.vanniktech.emoji.emoji.Emoji;
 
 final class EmojiSpan extends DynamicDrawableSpan {
-    private final float size;
-    private final Context context;
-    private final Emoji emoji;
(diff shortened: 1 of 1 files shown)


vanniktech (approved):
That makes sense! Can you try to keep the format?

cc @mario @rubengees

mario:
I'll review as soon as formatting is cleared

denjoo:
> I'll review as soon as formatting is cleared

I'm using the default formatter from Android Studio. Can you send me some information. How to format the code (xml, gradle)?

vanniktech (approved):

mario:
Looks decent to me.

mario (approved):

Issue #473, opened 2020-09-20
Just want to open popup without edittext usage


mihir0109:
Is there any possibility to use as bottomsheet or popup view? I don't have edittext.. I just want to open to select emoji on any button click.. and then close it.

vanniktech:
Not yet. Do you want to contribute this?

lukesleeman:
The project I'm currently working on also has a need for something like this.  One approach that may work for us is to just use `EmojiView` directly.

At the moment you can't use it in xml as it doesn't have the appropriate constructors, though once you get over that it basically works, and you can can integrate it in your UI however you see fit.

@mihir0109 would embedding `EmojiView` in your UI be appropriate for your usecase?  Or do you really need the "Keyboard like" functionality that this library provides via its popup?

Perhaps a possible way forward with this issue would be a PR which makes `EmojiView` easier to use by its self?

tobiasKaminsky:
We have a similar need.
For me it would be perfect, if it opens "keyboard like" popup when clicking on a button and I can handle result via OnEmojiClickListener.

mario:
Should be fairly trivial to implement, especially as you can pass the height of the view to the builder already. Looking forward to PRs.

lukesleeman:
Yeah, I might see what I can do about refactoring `EmojiView` to be a bit more like a "standard" android view.  At the moment you need a builder to construct one, and therefore you can't add it in as part of an xml layout.  There is also no documentation or examples of its use, so you really have to go digging through the codebase to figure it out.

KKTadena:
Found an issue using EmojiView:

1. Recent tab doesn't update
2. OnLongClick (to change some emoji color) doesn't work 
(EmojiView no default listener for onLongClick, setting null params for onLongClick crashes)

ps. hope we can hide the back button too

or there's just something I missed.

vanniktech:
Feel free to test this out in the latest master commit

vanniktech:
The custom view showcase in the sample app might be useful too, this one also only allows entering emojis

PR #496, opened 2021-05-17
Improved issue #473

xana-rahmani:
UseableEmojiView has been created to address issue #473 and aims to
 implement a version of EmojiView which is usable in xml files as an independent unit.
 it's possible that this implementation might not be the best solution for many cases, but it acts
 as a basic point for other developers to modify it corresponding to their needs.


vanniktech:
Thanks for your work but I'd imagine something that's not duplicated, instead something that's shared and can be used both ways.

vanniktech:
This is the end result: https://github.com/vanniktech/Emoji/pull/629

Issue #144, opened 2017-06-23
In place Dialog when typing in :emoji-alias: into EmojiEditText


vanniktech:
This would be optional and configurable on the EmojiEditText and could look something like this (taken from Slack):

![screenshot_20170623-150502](https://user-images.githubusercontent.com/5759366/27483398-7d6848f2-5825-11e7-85ab-fa5fcffd1d3b.png)

Relates to #102

rubengees:
The dataset the `generator` module uses, provides this information.
:grin: for example has the shortname `:grinning:` (This seems to differ a bit from other services, GitHub uses `:grin:` for that).
I can extend the `generator` to include this information into the `Emoji` object if you like.

vanniktech:
That would be handy as a first step, yes. Although I think some emojis have more than one aliases. So maybe `Set<String>` or `String[]` is the way to go.

rubengees:
Yes, the dataset provides those as `alias_alternatives` I think. There are also `keywords`, I will examine later what would make sense in this case and report back.

runegame:
When I send an emoji from the web, I receive it in this way (see image) is there any way to see the emoji?
![image](https://user-images.githubusercontent.com/3906550/32744355-9cc0197a-c885-11e7-885a-1b44877298ce.png)

vanniktech:
Don't hijack issues please.

vanniktech:
@rubengees have you had time to getting around including the aliases / keywords from the website?

rubengees:
@vanniktech Sorry, totally forgot about that!

I have pushed [a commit](https://github.com/rubengees/Emoji/commit/6504040219636433b00a74312d8bc534cb17e769) onto my fork with changes that implement this.
I will wait with a PR though, since we can't rerun the `generator` yet due to EmojiOne not having updated their `emoji.json` file yet...

vanniktech:
Oh right. That's still missing.

vanniktech:
Might happen or not. Will close for now.

vanniktech:
I really want this. @rubengees what's the easiest way we can get the tags? Once we have them we can build some UI.

rubengees:
I can do a PR for the tags I think.

vanniktech:
Much appreciated!

mario:
Where are we with this? :D

vanniktech:
We've got the text, no dialog yet.

PR #563, opened 2022-04-04
Emoji search icon next to categories, which allows to search emoji by name (english names only)

vanniktech:
Nothing implemented yet. How should we do this? WhatsApp does it this way:

Screenshot 2022-04-04 at 18 11 41

There's a little search button in the bottom left corner. When clicking this opens:

Screenshot 2022-04-04 at 18 11 55

```
+ pretty nice
- pain in the ass to implement probably (to have that perfect alignment between your app, our search bar and the keyboard)
```

---

As far as I know, Telegram has no 'direct' search: 

Screenshot 2022-04-04 at 18 14 00

What they do have is the special `:<letter>` thing when typing something though:

Screenshot 2022-04-04 at 18 15 38

This is tracked as https://github.com/vanniktech/Emoji/issues/144 and I'd like to have both options:

- both are cool
- when using force emojis only you would not be able to use the `:<letter>` trick
- to support Emoji Input usage without the keyboard: https://github.com/vanniktech/Emoji/issues/473

---

Currently with these changes our EmojiView looks like this:

Screenshot 2022-04-04 at 18 18 07

To try avoiding the keyboard hassle, I'd suggest opening a simple dialog when clicking on 🔍  where you can enter some text, it'll search Emojis and it'll list them. Any objections or better ideas?


vanniktech:
https://user-images.githubusercontent.com/5759366/165369615-74bd780a-8bd8-4d09-ac50-b29a96c5c752.mp4

vanniktech:
Merging this now, feel free to review though and I can make follow ups

PR #630, opened 2022-04-27
In place dialog when typing : in an EditText constructed from EmojiPopup.

### LINKED ISSUE #144: In place Dialog when typing in :emoji-alias: into EmojiEditText

vanniktech:
Fixes #144


vanniktech:
https://user-images.githubusercontent.com/5759366/165534230-b226e8e8-b418-4834-a7d0-94c9962bf431.mp4

PR #632, opened 2022-04-27
Emoji Search: Highlight which part of the shortcode matches the query.

vanniktech:


vanniktech:
Much better

android_screenshot_1651094805

PR #723, opened 2022-05-04
emoji-ios is now Kotlin only.

vanniktech:


hvar90:
hi,
a questionn 
 does it mean is not possible to use library version 10 on java app?

vanniktech:
Yeah

hvar90:
> Yeah

but i think with android Studio is possible mix kotlin libraries with java code, i have used kotlin libraries with my java/android project

hvar90:
many people use java code still, the library would be useful for both not only for one only

vanniktech:
Kotlin is completely interoperable with Java

PR #780, opened 2022-05-11
Start with Kotlin Multiplatform (android only for now)

vanniktech:


vanniktech:
@gabrielittner Is there anything that's missing? I'm trying to test things out by publishing to my local repository using `./gradlew publishToMavenLocal` and then using `mavenLocal()` with `0.11.0-SNAPSHOT` in a project of mine, however I'm getting:

```
> No matching variant of com.vanniktech:emoji:0.11.0-SNAPSHOT was found. The consumer was configured to find a runtime of a component, preferably optimized for Android, as well as attribute 'com.android.build.api.attributes.BuildTypeAttr' with value 'debug', attribute 'com.android.build.api.attributes.AgpVersionAttr' with value '7.2.0', attribute 'org.jetbrains.kotlin.platform.type' with value 'androidJvm' but:
    - Variant 'metadataApiElements' capability com.vanniktech:emoji:0.11.0-SNAPSHOT:
        - Incompatible because this component declares a usage of 'kotlin-metadata' of a component, as well as attribute 'org.jetbrains.kotlin.platform.type' with value 'common' and the consumer needed a runtime of a component, as well as attribute 'org.jetbrains.kotlin.platform.type' with value 'androidJvm'
        - Other compatible attributes:
            - Doesn't say anything about com.android.build.api.attributes.AgpVersionAttr (required '7.2.0')
            - Doesn't say anything about com.android.build.api.attributes.BuildTypeAttr (required 'debug')
            - Doesn't say anything about its target Java environment (preferred optimized for Android)
```

There are 2 commits here, one with the new base plugin and one without it (deprecated one). Neither one are working. Is there something obvious I'm doing wrong?

gabrielittner:
What version of AGP/Kotlin is that other project where you try to use the snapshot using?

vanniktech:
AGP 7.2.0 + Kotlin 1.6.21

vanniktech:
@gabrielittner publishing this with AGP 7.2.0 does not work, downgraded to 7.0.4 and it does now

PR #1039, opened 2023-12-09
Remove `stable` package in favor of native stable sort

boidolr:
For Node.JS since version 12 sort is already stable, meaning anything released after 2019 is fine (compare https://nodejs.org/en/about/previous-releases).


vanniktech (approved):

vanniktech:
Verified locally, thanks!

